Issue #3, opened 2021-03-31
Clean Production Spark configs


diocas:
Create a SWAN chart that has no Spark configuration (the hooks in the config file). 
Deploying the chart should provide a working SWAN, just like in production, but without Spark.

Issue #7, opened 2021-03-31
Centralize configuration values


diocas:
Ensure a consistent and single way of setting values (i.e., the auth values should be set in a single place/only once and imported everywhere where they are needed, like environment variables and config file); Investigate if some options in the current swan.*values.yaml files are needed (i.e., the impression I have is that we have auth config in multiple places and maybe they are not needed).

We should also take the opportunity to ensure that all configuration options of SWAN are easily changeable.

PR #16, opened 2021-04-29
Sparkconf

rccern:
#3


rccern:
There also some changes for authentication (swan_config.py has no "authentication section anymore) #7

PR #17, opened 2021-04-29
Configuration improvements

### LINKED ISSUE #6: JH dynamic configuration
Ensure the JH configuration file (`swan-chart/files/swan_config.py`) is configurable and is aligned with production. 
This means that all the configuration options need to be the same as we have in Puppet and that we do not have any hardcoded values unless they won't need to change (they should be changeable by setting them in values.yml).

 It should be possible to add extra python code to the configuration by setting it via values.yaml (or by importing a file?).
This will be necessary to set the Spark configurations in the production yaml file/chart.

Maybe we can start by analysing what options we need and discuss which ones should be configurable or not.

### LINKED ISSUE #7: Centralize configuration values

diocas:
Centralize all configs in `values.yaml` and use a separate `swan.dev.values.yaml` file with _only_ the differences for dev environments. @rccern Please check if we should have different configs by default. Also check if there's anything that we no longer use.

This makes configuration easier.
Closes #6 
Closes #7 

The next step will be removing some of these, the ones only related to CERN (i.e the authentication, the spark), and put them in a separate chart called `swan-cern`.

Also, there are some options that we need to use templates (don't know how to do that). For example, in the SSO redirect url and the ingress hostname need to be the same.

FYI @ebocchi


diocas:
@rccern feel free to push directly to this branch any fix you need. Can you also do a proper test?

rccern:
Hi, let me know once i can test this, for now helm install fails

rccern (approved):

PR #22, opened 2021-05-28
Support eos||cvmfs csi or bind mount

### LINKED ISSUE #8: Umbrella chart
Create the `swan` chart that automatically installs EOS and CVMFS.
Set values.yaml with the default options.

### LINKED ISSUE #9: Push SWAN chart to Harbor
Configure harbor (make it public) and push the SWAN charts there. Update documentation on how to deploy SWAN without having to clone the charts repo.

### LINKED ISSUE #11: Configurable EOS+CVMFS
Change the umbrella to allow the deployment of CSI or ScienceBox EOS+CVMFS charts.

ebocchi:
Support choice of CSI driver or bind mount for EOS and CVMFS.
Problems:
- we should decide a default and provide `values.yaml` accordingly
- `values.yaml` is not committed yet (need to decide the default and clean-up accordingly)
- It is hard to keep the ability to choose cvmfs (csi VS non-csi) for the mount in the hub. This is declared in `values.yaml`, which by designed is not templated. To be seen if we can work-around that.

This PR:
Closes #8 
Closes #9 
Closes #11


swan/Chart.lock
```diff
@@ -1,9 +0,0 @@
-dependencies:
```
diocas, line 1: Why are we deleting the Lock? Lock files are important to ensure consistency.

swan/templates/config.yaml
```diff
@@ -6,4 +6,311 @@ metadata:
...
+                "command": cull_cmd,
+            }
+        )
+
+    # init lists for volumes and volume_mounts
+    c.SwanKubeSpawner.volumes = []
+    c.SwanKubeSpawner.volume_mounts = []
+
+    # add EOS to notebook pods
+    {{- if .Values.eosClient.csiDriver }}
```
diocas, line 223: If I understood correctly, this is the only reason why you brought the configs directly to this file.
But can't we just ensure the same result by retrieving the config within the file and doing a normal `if` `else`? (need to check if you can retrieve _any_ value of the config, even those outside of JH).

I would prefer that route, because this makes it inconsistent with the rest of files that get loaded.

diocas: Also, can't we just make things uniform and use `/eos` or `/var/eos` in both?

diocas (commented):
Some questions before approving.

Reply to your points/problems:
- I guess the default will be non-csi.
- Aren't the defaults already in `values.yaml`?
- Did not understand your last point, sorry.

swan/templates/config.yaml
diocas: After a quick check, I think we cannot get these values, unless they are under `hub`, for example.

swan/Chart.lock
ebocchi: Maybe you are used to `helm dependency build`.
I typically use `helm dependency update`, which creates the `Charts.lock` according to the dependencies expressed in `Charts.yaml`.

I am not aware of differences between the two, besides personal preferences.
It might be that the `.lock` goes out of sync with the `.yaml` in case dependencies (or their versions) are updated in the `.yaml`. Weird that they use this sort of double source of truth....

All in all, tell me what you prefer to use and I will make it work ;)

swan/templates/config.yaml
ebocchi: Correct: The file was moved to templates so that be able to render it.
It is on personal preference:
- My option gives k8s a clean, ready-to-use chart according to the choice made and the whole logic is there
-  Your option shifts the choice to Python, which would be equally good.

If my option is inconsistent with other things already implemented (now it is clear to me why you have `/files`), I will move it back and put the logic in python.
To be checked exactly how we can propagate these values to the python script.
Will keep you posted.

ebocchi: And for where to mount eos, yes, we can use the same path on the host but I would still prefer to keep the switch for csi VS non-csi. Just in case of future breaking changes and for consistency with cvmfs.

swan/templates/config.yaml
```diff
@@ -6,4 +6,311 @@ metadata:
...
+        )
+    )
+    c.SwanKubeSpawner.volume_mounts.append(
+        client.V1VolumeMount(
+            name='eos',
+            mount_path='/eos',
+            mount_propagation='HostToContainer'
+        )
+    )
+    {{- else }}
```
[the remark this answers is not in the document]
ebocchi: Yes.

It is also true that we should also check the boolean value of `local_home` in that case.
However, you might want to have eos mounted with `local_home=true`. But you must have eos with `local_home=false`.

ebocchi:
Thanks for the comments. Will provide fixes (from Monday)

About the rest:
- _I guess the default will be non-csi._ ==> OK
- _Aren't the defaults already in values.yaml?_ ==> Yes, but according to csi VS non-csi the params change slightly
- _Did not understand your last point, sorry._ ==> You specify in `values.yaml` an extra volume on the `jupyterhub` to retrieve the spark config from cvmfs. This mount should also depend on csi VS non-csi but one cannot do `if-else` in values.yaml. Hence, I do not know how to specify persistentVolumeClaim (for csi) or hostPath (non-csi) alternatively in the list of mounts for `jupyterhub`. Will scratch my head further. Alternatives are welcome.

swan/templates/config.yaml
ebocchi: Answering to your _I think we cannot get these values, unless they are under hub, or custom_, I think we should put them into `hub.extraEnv`.
How about:
- `swan_config.py` goes back to `/files`
- if/else logic goes to `swan_config.py`
- pass envvars via `hub.extraEnv` to tell python what to do
- define the envvars in `values.yaml` by anchoring the values of {eos, cvmfs}Client.daemonSet/csiDriver
?

swan/Chart.lock
diocas: I don't _really_ care which one we use, but I have the experience from other places (like npm, in the javascript world), in which the lock file is important to ensure that we pull always the same dependencies, thus guaranteeing that things work. 
Here our charts don't have indirect dependencies, so when you update, the lock only contains direct dependencies, so it's not that important. It was more a matter of principle, but I have no idea if publishing the chart relies somewhat on these lock files or not.

swan/templates/config.yaml
diocas: I slept on this... `extraEnv` is for Environment variables, so not a good place (and I don't want to use env vars in the if/else).
I think it would make sense to put these values under `singleuser`, as these options are used by those pods only (not really in our production, but that's another story). So, something like `singleuser.cvmfs.{}` and `singleuser.eos.{}`. We might want as well to keep only that place as the source of truth and remove all the anchors? This means using that in the Chart dependencies as well. What do you think? (let me just test that those values are available in the Python realm).

I have another observation: we need to add an extra option when selecting the deployment type for EOS/CVMFS.. Why? You can create clusters that come with the CSI components out of the box, so we might want to set both the csi and the daemonset to False. But we still need to add them in the Python config, so in the `if` we should do something like `if csi or forceCsi`.

It's true it's on a personal preference, but if we have this working under `singleuser`, do you like it or not really?

diocas:
> You specify in values.yaml an extra volume on the jupyterhub to retrieve the spark config from cvmfs.

This is only in our production, so I guess that putting some comments in there warning about this dependency should be fine, no? It's not _elegant_, but as a short term/easy fix solution should be ok. If was for external usage, then yes, that would need a fix.

swan/templates/config.yaml
diocas: > let me just test that those values are available in the Python realm

Looks like it works

swan/Chart.lock
ebocchi: Reverted with 0560d1366b818ba5472ba729a120531961813c01

swan/templates/config.yaml
ebocchi: Yes, you are totally right about the 3rd option. The current solution clearly comes from my biased view of having nothing in the cluster (which is not the case in OpenStack).

Working on the singleuser.cvmfs.{} and singleuser.eos.{}

ebocchi: Globals won't work :/
Imlemented the three configuration options (daemonSet, CSI deploy, CSI already there) as per https://github.com/swan-cern/swan-charts/commit/e9e231b93e4130a33afe535cca070a2037526b90

ebocchi:
See if you better like this now.
Sadly, globals are not an option here but you get the three configuration options and both CVMFS and EOS are passed as _features_ of singleuser.

swan/files/swan_config.py
```diff
@@ -233,34 +220,95 @@ def swan_pod_hook(spawner, pod):
...
+    c.SwanKubeSpawner.volumes.append(
+        client.V1Volume(
+            name='eos',
+            host_path=client.V1HostPathVolumeSource(
+                path='/var/eos'
+            )
+        ),
+    )
+else:
+    # No access to EOS provided
```
diocas, line 267: Maybe log something here? If easy to do...

swan/files/swan_config.py
```diff
@@ -233,34 +220,95 @@ def swan_pod_hook(spawner, pod):
...
+        )
+        c.SwanKubeSpawner.volume_mounts.append(
+            client.V1VolumeMount(
+                name='cvmfs-'+cvmfs_repo_id,
+                mount_path='/cvmfs/'+cvmfs_repo_path['mount'],
+                read_only=True
+            )
+        )
+else:
+    # No access to CVMFS provided -- Nothing will work.
```
diocas, line 310: As above, some log message.

swan/values.yaml
```diff
@@ -1,8 +1,56 @@
...
+# - Deploy a DaemonSet pod with the CVMFS client running inside
+# - Chunks from upstream are fetched connecting to the CERN Stratum 1 server
+#
+# Warning:
+# - It is discouraged to enable more than one at once.
+# - By setting all to false, access to EOS will not be possible
+#     and singleuser's session will not be able to start.
+#
 cvmfs:
+  deployDaemonSet: &cvmfsDeployDS true
```
diocas, line 23: Can you also update `swan-cern` and the dev values?

diocas (changes requested):
Just small changes, but I might help if you don't have time. Otherwise, it's perfect! Super clean.

swan/files/swan_config.py
ebocchi: Fixed by c4143ca0a26774558b344734e86c11b1b044f6e3

swan/values.yaml
ebocchi: Fixed by 36ab2639dd48fca47ad680239a7c0332c4ec16d1

swan/templates/storageclasses.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.cvmfsClient.csiDriver }}
+{{ if or .Values.cvmfs.deployCsiDriver .Values.cvmfs.useCsiDriver }}
 {{- range .Values.cvmfs.repositories }}
```
diocas, line 2: This needs fixing, right? It's no longer in `cvmfs.repositories` but `singleuser.cvmfs.repositories` right?

diocas: Edited your last commit to fix it. Shout if I'm wrong @ebocchi

diocas (approved):

diocas:
@rccern can you test again in our QA? We need to change our config file because the variables that deploy csi vs daemonset have changed. Can you take care of it?
If all is ok, let's push it to production (need to update that config file as well), but only after CSC.
Btw, I think that none of the config files have been added to teigi.

PR #23, opened 2021-05-31
Rcas dev

### LINKED ISSUE #21: fluentd as dependency of swan-cern
Make it a dependency to be able to deploy all together in a single command.
Put the config file as an independent file to avoid polluting the config yaml.

If possible, in a different namespace? (I read helm allows it, but I don't know if it's true)

rccern:
Closes #21


swan-cern/Chart.yaml
```diff
@@ -10,4 +10,7 @@ description: The chart to deploy SWAN at CERN
 dependencies:
   - name: swan
     version: 0.0.1
-    repository: file://../swan
\ No newline at end of file
+    repository: file://../swan
+  - name: swan-fluentd
```
diocas, line 14: I think there's a better way of doing this. When I checked the Helm documentation on subcharts, I got the impression that was enough to create a new chart inside `charts/`. I think this would remove the need of declaration in dependencies and would keep the folders organized.
Can you check?

diocas (changes requested):
I think this needs a cleanup: the tgz file should be removed, requirements should no longer exist and the dependencies should be declared inside Chart.yaml. Basically, this should be cleaned the same way we cleaned the other charts.
I also have a comment on how to make this a dependency of `swan-cern`. Please check if my assumption is right.

We had discussed putting the config file outside of values.yaml, but I guess this doesn't matter much and can stay as is.

diocas (approved):
Discussed in Mattermost and decided to make it even better and directly use fluentd as dependency of `swan-cern` without the need of a subchart.

PR #24, opened 2021-05-31
Update QA

### LINKED ISSUE #10: SWAN@CERN
Use the `swan` chart (ideally it should just be a different version of values.yaml) to configure SWAN production @CERN. 

It should set the extra python code that configures Spark, as made possible by #3 and #6, and the auth configs. Everything else should come out-of-the-box from the swan chart.

### LINKED ISSUE #12: Configure EOS to use oAuth
Remove everything related to kerberos from the chart and configure oAuth. To be seen if the CM-provided EOS chart works. If not we need to replace it with our own chart (#11 addresses this).

diocas:
Merging our dev branch into qa.

This PR cleans and organizes our charts according to the good practices of Helm. While we don't push to Harbor, we rely on local dependencies.

Now we have a `swan` chart and a `swan-cern`, our production deployment chart that inherits from the first one and adds Spark, Fluentd, Kerberos for EOS and many configuration options.

We also align the config with production (authenticator, latest lcg, etc).

The missing updates (pushing to harbor and the EOS/CVMFS dependencies will come after).

Issues:
* Implements part of #8
* Closes #10
* Closes #12
  * It was agreed to keep Kerberos in our Production while we don't finalize the testing of EOS with oAuth. But we can consider this task closed.


diocas:
@ebocchi Please check. I know you are doing some changes, but I guess there is no major issue in these commits that prevents us from merging.

rccern (approved):
LGTM as we are already running it in the qa cluster

ebocchi (approved):
Looks good to me!

PR #33, opened 2021-08-16
Remove need to have a bastion host for deployments

rccern:


diocas:
@rccern this seems a bit risky... Imagine I'm in aiadm and cd into my ~/public folder before calling the script? Then the secrets will be there. We eventually remove them, but if something fails, we don't... I would at minimum print some warning message asking the user to ensure that he's in a secure location and acknowledging before continuing.

diocas (changes requested):
@rccern this seems a bit risky... Imagine I'm in aiadm and cd into my ~/public folder before calling the script? Then the secrets will be there. We eventually remove them, but if something fails, we don't... I would at minimum print some warning message asking the user to ensure that he's in a secure location and acknowledging before continuing.

rccern:
In case you cd accidentally into your public folder the script won't run already, as it checks to be in the git folder. 
Anyway, I added as an additional measure added a trap to delete the files.

diocas:
We are supposed to git clone and run the script right? So what I said is, imagine that before you did the clone, you cd into some public or semi public path. Ofc you would run from inside the git clone, but that can be anywhere.

rccern:
I don't think it's up to the script to make sure that you are not in a public folder... tbag doesn't do that for example. Ofc it assumes that the user knows he's pulling a secret, which is assumed also in this case

rccern:
I fixed the deletion behaviour

diocas (approved):
Approved and going to merge.
And I disagree with what you said. In tbag you know that you are handling with secrets and you explicitly store them in a file, here you don't necessarily know because you don't say it anywhere (unless you read the script). That's why I asked to inform the user that secrets were being stored locally temporarily. But I think your approach is equally good.

PR #34, opened 2021-08-23
Fixselinux

rccern:


swan-cern/files/swan_config_cern.py
```diff
@@ -187,6 +197,23 @@ def _init_eos_containers(self, eos_secret_name):
         # assigning pod spec containers
         self.pod.spec.containers = pod_spec_containers
 
+    def _gpu_enabled(self):
```
diocas, line 200: This method already comes with the `swan` chart (the file will be evaluated before), so I don't think it's required. Did you check?

swan-cern/files/swan_config_cern.py
```diff
@@ -24,6 +24,16 @@ def get_swan_user_pod(self):
         # init user containers (notebook and side-container)
         self._init_eos_containers(eos_secret_name)
 
+        if self._gpu_enabled():
```
diocas, line 27: I still believe we need to rethink the gpu logic. If it belongs here or in the spawner, we need to see.
But, this might need to be optional as well.

swan-cern/files/swan_config_cern.py
rccern: right,fixed

swan-cern/files/swan_config_cern.py
rccern: ok, but i suggest we merge this and revisit later, what do you think?

diocas: yes

diocas:
Please rebase (remove the changes in the deploy script) and it's ready to merge.

PR #36, opened 2021-09-06
Use new fusex@0.1.0 chart

ebocchi:
No other changes than using a more recent fusex chart and bumping swan chart at 0.0.5
This is used by ScienceBox and should not interfere with swan-cern.


ebocchi:
Let me know if ok so that I can tag the new chart

diocas (approved):
Merging. Go ahead with the tag!

PR #38, opened 2021-09-16
change to swan chart for upstream 1.1

rccern:


diocas (dismissed):
Looks fine, please merge.
I guess you'll do the update to swan-cern once this is merged.
The need to put the eos/cvmfs changes in custom, is it any limitation from the new version?
Also, with this version can we already take this out? https://github.com/swan-cern/swan-charts/blob/master/swan/values.yaml#L191 (we also have it in swan-cern).

rccern:
Good catch, the config dir was fixed in 1.0, I'll remove that line. For swan-cern  I will open another PR

PR #63, opened 2022-04-08
customize gpu error message

rccern:


swan/files/swan_config.py
```diff
@@ -80,7 +80,7 @@ def _gpu_enabled(self):
         lcg_rel = self.spawner.user_options[self.spawner.lcg_rel_field]
 
         if "cu" in lcg_rel and "swan-gpu" not in user_roles:
-            raise ValueError("Access to GPUs is not granted; please contact swan-admins@cern.ch")
```
diocas, line 83: I don't think this check should exist in swan, only in swan-cern. We don't have groups in sciencebox, which means this would always be false. Maybe allow it by default (we can always remove it from the stack if not available)?

diocas (changes requested):

swan/files/swan_config.py
rccern: ok, i'll just move to the file where the pod hook  is implemented in swan-cern

diocas (approved):

PR #66, opened 2022-05-17
Remove EOS Kerberos Authentication

krishnan-r:
TODO
- [ ] convert the  sidecar to an `initContainer` to fix spark cloud containers configuration

To be merged once we upgrade eosxd and oauth is stable


diocas:
Can we remove our `swanculler` as well? It was needed to renew the EOS tickets, which is no longer needed. (we might need extra stuff - like hooks - in the future, but we should do that with the upstream version).
Removing the culler means removing its config in our config.py, delete the python module from the  jupyterhub-extensions repo, and finally removing its installation from the jh image.

PR #67, opened 2022-05-18
set eosxd update strategy to ondelete

rccern:
this allows us to update eos more gently: we first cordon the node, wait for users to go to another node and then update the eos version.


diocas (commented):
@ebocchi I guess we would benefit from having this in `fusex` as well.

@rccern I think it's worth it to put some docs explaining it

diocas:
Assumed that the notes would be there for the ops, but maybe an simple explanation here to prevent people from deleting it in the future or wondering why it's being set.
btw, I think you pushed some fluentd stuff here by mistake.

rccern:
should be fine now. fluentd stuff should go away once we rebase

diocas (approved):

PR #73, opened 2022-07-19
Cleanup spark NodePort services

krishnan-r:
- To allow spark/dask workers to reach the driver running in a user pod from outside the cluster, we open ports using `type: NodePort` services

- We have a limited number of ports available for `type: NodePort` kubernetes services. The available range is specified through the `--service-node-port-range` [flag](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) (default: 30000-32767) when creating the cluster. This currently only support ~150 users with 18 ports open for each user session/pod.

This change cleans up these services when the spawner stops the user server, to prevent them from accumulating over time.

Related: #64 fixed issues caused by too many environment variables due to many services in the namespace, causing timeouts in EOS daemon trying to read the environment for auth tokens.

### DIFF
--- a/swan-cern/files/swan_spark_config.py
+++ b/swan-cern/files/swan_spark_config.py
@@ -347,7 +347,20 @@ def spark_modify_pod_hook(spawner, pod):
     spark_pod_hook_handler = SwanSparkPodHookHandler(spawner, pod)
     return spark_pod_hook_handler.get_swan_user_pod()
 
-# Get configuration parameters from environment variables
-# swan_container_namespace = os.environ.get('POD_NAMESPACE', 'default')
+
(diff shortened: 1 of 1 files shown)


etejedor (approved):
Thanks! We'll need to adjust this again for the HTCondor integration and create&delete more NodePort services if the user selects that they want to use HTCondor in their session. We will need to decide how many ports we allocate for HTCondor cluster connections -- there can be multiple of them in the same session, as for Spark, since the user will be able to create multiple SwanHTCondorClusters from the UI.

As per the limitation of 150 sessions with 18 ports each, I guess we are fine for now. It will only become a problem if our users start using more Spark / HTCondor.

PR #80, opened 2022-08-12
Enable CERN HTCondor configuration flag

krishnan-r:
- Aligns with puppet nodes to create `/etc/condor` from the batch service configuration
- Sets the environment variable to toggle the [flag](https://github.com/swan-cern/systemuser-image/blob/effdb9edb0262ba75ff94e19b914049460dd3e92/systemuser.sh#L216) in the image.


etejedor (approved):
Thanks for enabling this! Let's have this always on, even if in the future the user does not choose to use HTCondor in distributed (Dask) mode, because they could still want to submit batch jobs via command line.

PR #85, opened 2022-09-21
Enable Liveness Probe for Hub Pod

krishnan-r:
- Waits for 5 minutes for hub to start up.
- Restarts hub container if probe fails for 20 * 15 = 5 minutes

- I have increased the timeouts from the [upstream defaults](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/5c0d0c5c74bc97bf403844ced6b8e79cc089498d/jupyterhub/values.yaml#L118) (adjusting other values to restart in 5 minutes) to ensure we don't restart the hub unless it is really stuck.

### DIFF
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -125,7 +125,11 @@ jupyterhub:
         cpu: 200m
         memory: 512Mi
     livenessProbe:
-      enabled: false
+      enabled: true
+      initialDelaySeconds: 300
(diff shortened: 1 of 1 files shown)


etejedor:
Thanks @krishnan-r , is this to try to self heal the hub in those situations we observed when it gets stuck, since restarting the hub pod is the only solution we have found so far?

PR #86, opened 2022-10-11
Workflow to automate releasing charts

### LINKED ISSUE #30: Build/publish using github actions
To make it uniform with the other repos and integrate with PRs.
To fix the current ci which looks like it's pushing all charts instead of just the tagged one.

Also, to uniform with the other repos (and allow grouping in git clients), tags should be `swan/vX.X.X` or `swan-cern/vX.X.X`.

Useful action:
https://github.com/marketplace/actions/helm3-push-action

ps: need to create a dedicated service account for publishing.

krishnan-r:
This PR adds Github Actions workflows to lint both the charts and tag and publish charts to harbor on demand.

- The workflow lints PRs using `helm lint` 

- When the 'Release Charts' workflow is manually triggered it
  - Fails if not on master branch
  - Lints charts
  - Checks if any of the charts have changes since the last tag, if no changes fails
  - If `swan/` has changes, updates Chart.yaml/version, creates tag, publishes chart to harbor, and creates a github release with the commit log of swan/ as release notes.
  - If swan/ is tagged, updates `swan-cern` with latest version of `swan/`
  - Repeats the tagging for `swan-cern` if it has changes.
  - If there are no changes in the base chart since last tag, it skips tagging a new version of the base chart and tags only the sub-chart

For an example of how it works, see the example runs in fork: https://github.com/krishnan-r/swan-charts/actions/runs/3291033601 and release, tag, and commits made by the workflow on master branch of that repo.
image

# Secrets
The workflow requires following secrets to be set on the repository/organization
- `WORKFLOW_ACCESS_TOKEN` A GitHub PAT with admin permissions to push tags, and create commits on master
- `HELM_REPO_USERNAME` and `HELM_REPO_PASSWORD` credentials of a Harbor robot account with permissions to list and push helm charts

#### TODO
- [x] bumpversion.cfg needs search regex as it will bump any dependencies if it has the same version as the chart. It currently does a naive find/replace on Chart.yaml. This also appears to be the case for other extensions
- [x] Update the Chart.yaml base URLs changed for testing before merging
- [ ] Configure secrets on repository

Fixes #30


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Temporarily change swan-cern to use local file://../swan
+        uses: mikefarah/yq@1f0881fb5faf371694bfa108753cda0b824f5037 # v4.28.2
+        with:
+          cmd: yq -i '(.dependencies[] | select(.name == "swan")).repository = "file://../swan"' swan-cern/Chart.yaml
```
etejedor, line 22: This means that both swan and swan-cern will be cloned to a local filesystem, and we are telling swan-cern to use the local swan as dependency instead of the Harbor one, right?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - run: helm dependency update swan && helm lint swan && helm package swan
```
etejedor, line 23: IIUC here we:
1. Download dependencies and generate lock file
2. Check the sanity of the chart
3. Generate a local packaged chart (not yet pushed to Harbor)

.github/workflows/lint.yaml
krishnan-r: Yes, this allows us to submit PRs that change both `swan/` and `swan-cern/`, without tagging one and then changing the other. 

Otherwise, `helm lint` on `swan-cern/` would lint against the version of `swan/` in the registry and fail due to missing changes that are not yet tagged in `swan/`

.github/workflows/lint.yaml
krishnan-r: yes, `helm package` generates a tar file, but here the output is not really used, instead just tests that it packages correctly.

I borrowed this from the existing Gitlab CI pipeline https://github.com/swan-cern/swan-charts/blob/0c42fda15b943492387bd8cd0c60dcdc17f9628d/.gitlab-ci.yml#L17

.github/workflows/list-changed-charts.sh
```diff
@@ -0,0 +1,21 @@
...
+for CHART_NAME in "${CHARTS[@]}"; do
+
+    # Determine last tag on the chart
+    LAST_CHART_TAG="$(git tag --list --sort -version:refname ${CHART_NAME}@* | head -n 1)"
+
+    # Determine if there are any changes on the chart/ folder since last tag
+    HAS_CHANGED="$(git diff --quiet HEAD ${LAST_CHART_TAG} -- ./${CHART_NAME} || echo true)"
+
+    # print output
+    echo "${CHART_NAME}_has_changed=${HAS_CHANGED}"
```
etejedor, line 19: You could make this a conditional and just output e.g. `swan_has_changed` if the result of the diff is 1, else not print at all.

.github/workflows/list-changed-charts.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
etejedor, line 2: I would add a comment here like: `# Prints a line "${CHART_NAME}_has_changed" for every chart that has changed since last tag`

swan-cern/.bumpversion.cfg
```diff
@@ -0,0 +1,8 @@
+[bumpversion]
+current_version = 0.1.11
+commit = True
+tag = True
+tag_name = swan-cern@{new_version}
+message = Bump chart swan-cern/ version: {current_version} → {new_version}
+
+[bumpversion:file:Chart.yaml]
```
etejedor, line 8: As per the comment you made in the PR message, this looks for a string with {current_version} in Chart.yaml and replaces it with {new_version}, no matter where?

etejedor: Also, how is {new_version} filled? Does the value come from what the user selects in the UI (minor, major)?

etejedor:
Thanks for automatizing this @krishnan-r !

General question: if swan has changes and therefore a new version is tagged, we need to update the dependencies of swan-cern with that new version. The fact of changing the version of the swan dependency in swan-cern also needs to trigger a new tag for swan-cern?

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Fetch all tags and history to check for changes
+      - name: List charts with changes since last tag
+        id: diff
+        run: |
+          set -euxo pipefail
+          ./.github/workflows/list-changed-charts.sh >> $GITHUB_OUTPUT
+      - name: Fail if no charts have changes
```
etejedor, line 51: Do you need to fail in this case or just do nothing and show a message like: `There are no changes to be tagged, aborting.`

.github/workflows/list-changed-charts.sh
krishnan-r: The output of this script is piped to `>> $GITHUB_OUTPUT` to set an [output for the step](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) which is mapped to the [job output](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs). This needs output in key=value format

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,118 @@
+# Workflow to tag and publish swan/ and swan-cern/ charts in this repository
```
etejedor, line 1: This is the main workflow that orchestrates everything, right? I would say it explicitly here. And I would list all the steps it orchestrates, here you only mention the tagging and commiting of new versions to the github repos, but there's also check_branch, lint, list_changed_charts and tag charts (which also including pushing to harbor with the corresponding secrets) with the dependency update in between.

.github/workflows/lint.yaml
etejedor: But what you clone here is used later in some other steps of the orchestration workflow?

It seems that  the purpose of this workflow is purely checking the correctness of the charts. The changes made to the lock file by helm dependency update and the package generated by helm package are not used anywhere else. Is this correct?

swan-cern/.bumpversion.cfg
krishnan-r: Yes, this is a simple find/replace, and can potentially conflict if any dependency has the same `current_version`, it gets bumped as well. (This is especially possible in our other extensions, in packages.json files particularly, I'm surprised we haven't had problems yet.)

The [bump2version](https://pypi.org/project/bump2version/) tool has a `search=default:{current_version}` option which can be customized with a python template that matches only the areas we want in the file.

The `new_version` is calculated from the 2nd argument to bump2version command, `part` which is major/minor/patch/'other parts in SemVer', and is currently an input to the workflow in this pipeline

.github/workflows/release-charts.yaml
krishnan-r: The workflow will appear as passed, with a green tick in the Actions Tab, but both `tag_swan` and `tag_swan_cern` jobs would be skipped. 

This might be confusing if we are expecting a new tag to appear. Marking the workflow as failed better conveys that nothing happened.

.github/workflows/lint.yaml
etejedor: Ok thanks, here you answer about the package. In the end here you are just checking that the chain update - lint - package runs without problems, but you are not reusing the products of update and package anywhere else.

I would explain this (the ultimate purpose of the workflow) as a comment at the beginning. And perhaps instead of calling it just "lint" it should be called "check_charts" or "charts_sanity_test" or similar?

.github/workflows/list-changed-charts.sh
etejedor: Ok, it's just that if there are no changes, you emit "swan_has_changed=". That just assigns a null value to the key and it is fine?

swan-cern/.bumpversion.cfg
etejedor: Ok thanks, but how and when and by whom is the version we are bumping to specified? Is it the same as with the extensions, that I select on the UI which version (major, minor, patch) I want to bump to? Does that setting apply to both swan and swan-cern?

.github/workflows/tag-chart.yaml
```diff
@@ -0,0 +1,120 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      new_version: ${{ steps.version.outputs.new_version }}
+      new_tag: ${{ steps.version.outputs.new_tag }}
+      last_tag: ${{ steps.version.outputs.last_tag }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false # To use WORKFLOW_ACCESS_TOKEN instead of GITHUB_TOKEN in later steps
```
etejedor, line 51: Can you explain this line and its comment?

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,118 @@
...
+# - uses secrets.WORKFLOW_ACCESS_TOKEN (GitHub PAT with admin permissions on repostiory) to push commits and tags
```
etejedor, line 2: Typo repostiory

.github/workflows/lint.yaml
krishnan-r: GitHub Actions "Jobs" potentially run on separate runners and do not persist changes on working directory on other jobs

The repo is cloned with a clean working directory on each job using `actions/checkout` step.

So the changes made here are not used anywhere else in the workflow. Only the job result  is used to fail the workflow if lint fails.

.github/workflows/tag-chart.yaml
krishnan-r: Every workflow in GitHub actions is provided a token by default stored in `$GITHUB_TOKEN` which is a built in service account for every repository, actions made on the repo using this token appear as done by `github-actions`.

This token however does not have permissions in the repository to push to protected branches and other actions that require admin privileges. We need to update the version with a commit on master which is protected, and push a tag.

To overcome this, we have a GitHub Personal Access Token (PAT) configured in the swan-cern organisation level secrets, with the token of a user with admin privileges (Diogo) which is stored in the secret WORKFLOW_ACCESS_TOKEN.

The checkout step caches the credentials for git in the `.git/` folder. In a later step where we want to use `WORKFLOW_ACCESS_TOKEN`, the cached `GITHUB_TOKEN` takes precedence, so we need to disable this behaviour

swan-cern/.bumpversion.cfg
krishnan-r: Currently when triggering the workflow manually, the part to bump is an input, with a drop down to select  'major', 'minor', and 'patch'.

This applies to both `swan/` and `swan-cern/`. If you select major, both charts will be tagged with a new major version.

We can also provide two inputs for each chart separately, what do you suggest?

etejedor: It might be the case that the changes for swan deserve a major version change, but not for swan-cern. Moreover, since we agreed both charts can have different versions, I believe it would be better to have two variables, one for each.

krishnan-r: I have updated it now to include two dropdowns to pick major/minor/patch for each

.github/workflows/list-changed-charts.sh
krishnan-r: Fixed this, now it prints =false or =true, previously I was never checking for false

krishnan-r:
I also updated the PR to remove the GitLab CI pipeline, (which would otherwise push the chart twice, again on tag pushed by this workflow)

krishnan-r:
@etejedor I have updated the PR with two more commits fixing the TODOs, once the secrets are configured, it should be ready to merge

etejedor (approved):

PR #89, opened 2022-10-31
Include username in CI version bump commit message

krishnan-r:
Add username to commits created when releasing from CI

example commits: https://github.com/krishnan-r/swan-charts/commits/master

### DIFF
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -111,7 +111,7 @@ jobs:
           git config user.name github-actions
           git config user.email github-actions@github.com
           git add .
-          git commit -m "Update swan-cern to use swan@${{ needs.tag_swan.outputs.new_version }}"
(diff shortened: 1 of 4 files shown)


etejedor (approved):

diocas:
why?
(btw, our commit messages could be aligned with the other extensions, i.e `swan v0.1.13` for releases and maybe smth like `swan-cern: update swan to v0.1.13` for the auto update... just nitpicking)

etejedor:
> why?

I don't see any harm in keeping a trace of who triggered the workflow. Just like commits have authors so that we can know who did what, even if one could argue that those commits are also done "for the team". Another example: github adds information of who merged a given commit, together with who the author was. Information is good :)

diocas:
But we do keep trace already, as you said all actions are logged in github (together with their status of completion). I know who triggered all releases (maybe just the admins have access to this?).
But fine, I guess this makes it slightly easier to know, even if the commits are a bit too verbose to my taste.

diocas (approved):

PR #91, opened 2022-11-02
Fix EOS Kerberos Auth failing due to large KRB5CC

krishnan-r:
The size of the kerberos credentials cache is large for some users.
The curl command to replace the secret fails because the arguments are too long. So instead this passes the JSON body as a file

- Also adds a `swan_user: ` label to the secret aligning with the pod labels

### DIFF
--- a/swan-cern/files/private/cull_check_ticket.sh
+++ b/swan-cern/files/private/cull_check_ticket.sh
@@ -15,6 +15,27 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+# The request body with the token exceeds the bash argument size limits.
+# So use a file with 'curl --data'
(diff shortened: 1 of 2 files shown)


diocas (approved):
So you found the issue with the lack of authentication on EOS... This is great!
(great to push, even if we're planning to migrate to oauth rather sooner than later.

etejedor:
@diocas indeed this was the reason for many refresh requests to fail, but we saw recently a case where the renewal succeeded (the kerberos credentials were not big enough to make it fail) and yet the user reported problems accessing EOS. So let's still be en garde wrt this.

diocas:
Yes, I'm aware that EOS getting stuck is a completely different issue, unfortunately..

PR #109, opened 2022-11-17
Delete jupyterhub_form.html

etejedor:
To be discussed.


krishnan-r (approved):

etejedor:
After discussion with @diocas and @krishnan-r , it was agreed that this file can be deleted from the swan-charts repository.

PR #128, opened 2023-01-19
Configure URL to SWAN gallery as an environment variable

krishnan-r:
- This is used by notebook to render the link in the template
- Used by jupyterlab to load the iframe website

Used by https://github.com/swan-cern/systemuser-image/pull/95

### DIFF
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -97,6 +97,7 @@ jupyterhub:
       kubectl.kubernetes.io/default-container: notebook
     extraEnv:
       SWAN_DISABLE_NOTIFICATIONS: "true"
+      GALLERY_URL: "https://swan-gallery.web.cern.ch"
   ingress:
     enabled: true
     annotations:


diocas (approved):

swan/values.yaml
```diff
@@ -97,6 +97,7 @@ jupyterhub:
       kubectl.kubernetes.io/default-container: notebook
     extraEnv:
       SWAN_DISABLE_NOTIFICATIONS: "true"
+      GALLERY_URL: "https://swan-gallery.web.cern.ch"
```
etejedor, line 100: The fact that you need to set the variable here means that, in our current code, we checked if the variable existed in multiple places (e.g. [1](https://github.com/swan-cern/jupyterhub-image/blob/master/jupyterhub.d/jupyterhub_config/kubespawner.py#L26), [2](https://github.com/swan-cern/systemuser-image/blob/master/systemuser.sh#L271) and [3](https://github.com/swan-cern/jupyter-extensions/blob/master/SwanContents/swancontents/templates/tree.html#L23)) but it was actually never set? For example in [3](https://github.com/swan-cern/jupyter-extensions/blob/master/SwanContents/swancontents/templates/tree.html#L23) the default was being taken instead?

diocas: Can you clarify what you mean, @etejedor ? I'm not understanding.

diocas: Also, AFAIK, that jupyterhub config file is no longer used.. It's a leftover from other times

etejedor: `GALLERY_URL` is defined here in this commit, but in the current master we have code that checks for this variable already (see links I provided). This means the variable was never set in k8s, and we just relied on defaults (see 3).

diocas: Yes, ofc we relied in the defaults. This option is for ScienceBox (this was used in the past for Up2U, as we had a special gallery there). And by checking you mean "setting the config if a value is defined"? If so, yes. 
Also, as I said, the code from first link you sent is not used anywhere (this is smth that has been cleaned in Riccardo PR that is waiting for the changes for the Spark stuff).

diocas:
I'm closing this PR as a default exists in the help extension

PR #129, opened 2023-01-23
Remove user pod affinity based on LCG release

etejedor:
Setting a pod affinity based on the LCG release that the user selected leads to pods being clustered in nodes by the kubernetes scheduler. This, together with overcommitting of the nodes, can cause memory pressure in crowded nodes where multiple users use the memory of their sessions up to the limit.

By removing this affinity setting, we seek to rely instead on the default strategy of the k8s scheduler, which tries to assign pods to least allocated nodes. This will make it less likely, unless the cluster is close to full capacity, that one or more users push a node's memory to the limit.

Note that the affinity setting was done to favour warm CVMFS caches in nodes so users had a better experience. However, most of our current users select the default software stack, the NXCALS stack or the CUDA stack, which are all warmed up automatically by our cron jobs. The few users that select other stacks will pay some cost for warming up their cache, but the penalty in time is not that significant anyway, i.e. the gains largely outweigh the losses with this change.

### DIFF
--- a/swan-cern/values.yaml
+++ b/swan-cern/values.yaml
@@ -92,6 +92,10 @@ swan:
       extraEnv:
         # Enable HTCondor service configuration for CERN in the user image
         CERN_HTCONDOR: "true"   
+    scheduling:
+      userPods:
+        nodeAffinity:
+          matchNodePurpose: ignore
     hub:
       extraVolumeMounts:
         - name: swan-jh-cern
--- a/swan/files/swan_config.py
+++ b/swan/files/swan_config.py
@@ -29,9 +29,6 @@ class SwanPodHookHandler:
             pod_labels
         )
 
-        # init pod affinity
-        self.pod.spec.affinity = self._init_pod_affinity(pod_labels)
-
         if self._gpu_enabled():
             # currently no cern customisation required
             self.pod.spec.volumes.append(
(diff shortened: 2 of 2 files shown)


krishnan-r (dismissed):
Note for the future:

Removing this also enables the default from the upstream chart [scheduling.userPods.nodeAffinity.matchNodePurpose](https://z2jh.jupyter.org/en/stable/resources/reference.html#scheduling-userpods-nodeaffinity-matchnodepurpose) = `prefer` which sets a [nodeAffinity](https://github.dev/jupyterhub/zero-to-jupyterhub-k8s/blob/23bf6a109fe9dd8fdabdbbaaa4c7d8abaf751951/jupyterhub/files/hub/jupyterhub_config.py#L211)




```json
{
  "nodeAffinity": {
    "preferredDuringSchedulingIgnoredDuringExecution": [
      {
        "preference": {
          "matchExpressions": [
            {
              "key": "hub.jupyter.org/node-purpose",
              "operator": "In",
              "values": [
                "user"
              ]
            }
          ]
        },
        "weight": 100
      }
    ]
  }
}
```

But since we do not label the nodes, this would not have any impact. (We can also set it to scheduling.userPods.nodeAffinity.matchNodePurpose: `ignore` to remove this as well)

etejedor:
Thanks @krishnan-r for pointing out the new node affinity that is set by the upstream chart. I think I will add the `ignore` setting that removes the affinity, since it  means less checks to be done by the scheduler when filtering out nodes.

etejedor:
Added the setting to remove node affinity, please check @krishnan-r .

krishnan-r (approved):

PR #133, opened 2023-01-26
[Draft] Overwrite eosxd configmap to set hide-versions:0 on the individual mo…

etejedor:
…unts

The setting is necessary for versions to show up and be visible by users via the Checkpoints option in the menu, to revert back to a previous checkpoint.

This is a temporary fix to address two issues identified:
- The eosxd upstream chart does not allow anymore to set options in a global fuse.conf configuration file in the eosxd configmap. This needs to be discussed with the owners of the eosxd chart and potentially allow it again.
- Event if a global fuse.conf exists, EOS does not seem to honor it. Only if hide-options:0 is set for the individual mounts, such setting is taken into account. This needs to be discussed with the EOS team and identify a way to set global flags.


diocas (approved):

etejedor:
Closing, since the k8s documentation does not specify what happens when two objects of the same type, namespace and name are defined in different charts (i.e. we can't ensure the precedence of this configmap vs the eosxd configmap).

PR #134, opened 2023-01-30
Change resource settings for user pods, eosxd and cvmfs

etejedor:
This PR proposes a few changes in the resource specification of user, eosxd and cvmfs pods to:
- User pods: make the memory resource request dynamic, more precisely a fraction of the limit.
- eosxd and cvmfs: set a baseline memory request so that the k8s scheduler is aware from the start that such pods have some memory consumption.

### DIFF
--- a/swan-cern/values.yaml
+++ b/swan-cern/values.yaml
@@ -62,8 +62,6 @@ swan:
     useCsiDriver: &eosUseCSI true
   jupyterhub:
     singleuser:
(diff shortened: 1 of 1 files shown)


etejedor:
Note: the CVMFS settings will only be effective when updating to a recent tag of the cvmfs chart (0.0.7).

diocas (approved):

PR #139, opened 2023-02-25
Cleanup user secrets when a session stops

krishnan-r:
Deletes the eos and hadoop tokens secret from a hook in the spawner when a server stops.

### DIFF
--- a/swan-cern/files/swan_config_cern.py
+++ b/swan-cern/files/swan_config_cern.py
@@ -251,5 +251,15 @@ swan_container_namespace = os.environ.get('POD_NAMESPACE', 'default')
 
 c.SwanKubeSpawner.modify_pod_hook = swan_pod_hook_prod
 
+def swan_cern_post_stop_hook(spawner):
+    # Delete Kubernetes Secret storing eos kerberos ticket of the user
+    username = spawner.user.name
+    eos_secret_name = f"eos-tokens-{username}"
+    swan_container_namespace = os.environ.get('POD_NAMESPACE', 'default')
+    spawner.log.info('Deleting secret %s', eos_secret_name)
+    spawner.api.delete_namespaced_secret(eos_secret_name, swan_container_namespace)
+
+c.SwanKubeSpawner.post_stop_hook = swan_cern_post_stop_hook
+
 # Required for swan systemuser.sh
 c.SwanKubeSpawner.cmd = None
--- a/swan-cern/files/swan_spark_config.py
+++ b/swan-cern/files/swan_spark_config.py
@@ -348,19 +348,31 @@ def spark_modify_pod_hook(spawner, pod):
     return spark_pod_hook_handler.get_swan_user_pod()
 
 
-def spark_cleanup_services(spawner):
+def spark_post_stop_hook(spawner):
     """
     :param spawner: Swan Kubernetes Spawner
     :type spawner: swanspawner.SwanKubeSpawner
-    
     """
+    
+    # Call the parent hook defined in the swan_config_cern.py config file
+    # This function is assumed to be available as a global, because the config files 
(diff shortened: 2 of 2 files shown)


swan-cern/files/swan_spark_config.py
```diff
@@ -347,20 +347,32 @@ def spark_modify_pod_hook(spawner, pod):
...
+            spawner.api.delete_namespaced_secret(hadoop_secret_name, swan_container_namespace)
+
+def spark_post_stop_hook(spawner):
+   spark_post_stop_hook_handler = SwanSparkPostStopHookHandler()
+   spark_post_stop_hook_handler.post_stop_hook(spawner)
 
 c.SwanKubeSpawner.modify_pod_hook = spark_modify_pod_hook
-c.SwanKubeSpawner.post_stop_hook = spark_cleanup_services
\ No newline at end of file
+c.SwanKubeSpawner.post_stop_hook = spark_post_stop_hook
```
etejedor, line 378: This is supposed to overwrite the setting done in `swan_config_cern.py` for `SwanKubeSpawner.post_stop_hook`? Which one has precedence?

swan-cern/files/swan_config_cern.py
```diff
@@ -251,5 +251,21 @@ def swan_pod_hook_prod(spawner, pod):
 
 c.SwanKubeSpawner.modify_pod_hook = swan_pod_hook_prod
 
+class SwanPostStopHookHandler:
+    def post_stop_hook(self, spawner):
+         # Delete Kubernetes Secret storing eos tokens
```
etejedor, line 256: Is this the EOS token exchanged with SSO? Only EOS or also CERNBox? If both I'd suggest, `Delete Kubernetes secret storing exchanged tokens for EOS and CERNBox`.

swan-cern/files/swan_spark_config.py
```diff
@@ -347,20 +347,32 @@ def spark_modify_pod_hook(spawner, pod):
...
-    
-    """
-    spark_cluster = spawner.user_options[spawner.spark_cluster_field]
-    if spark_cluster and spark_cluster != 'none':
-        username = spawner.user.name
-        spark_ports_service = f"spark-ports-{username}"
-        spawner.log.info('Deleting service %s', spark_ports_service)
-        swan_container_namespace = os.environ.get('POD_NAMESPACE', 'default')
-        spawner.api.delete_namespaced_service(spark_ports_service, swan_container_namespace)
+class SwanSparkPostStopHookHandler(SwanPostStopHookHandler):
```
etejedor, line 350: Do we need this to be a class? If it's just for calling `super`, we can do that defining two functions and invoking one from the other. We don't use `self` to store any value anyway, it just adds boilerplate imo.

swan-cern/files/swan_spark_config.py
krishnan-r: This precedence is I think based on the alphabetic ordering of the files in this config folder [here](https://github.com/swan-cern/swan-charts/blob/1c3895446147849764c5393d984ad3ea6416934b/swan-cern/values.yaml#L87)

swan-cern/files/swan_config_cern.py
krishnan-r: OAuth tokens are downloaded directly from `/api/user` which also returns the `auth_state` stored in the jupyterhub database (custom handler implemented by [SwanHub](https://github.com/swan-cern/jupyterhub-extensions/blob/9d7e2e73992c29dcf5438e0ff7a92b1b4025a66d/SwanHub/swanhub/userapi_handler.py#L26)). 

The Kubernetes secret stores the EOS kerberos secret.

etejedor (approved):
LGTM thanks!

PR #154, opened 2023-05-05
auth: request openid scope for standard compliance

diocas:
Keycloak is going to be updated at CERN and without this scope calls to userinfo endpoint will fail.

AFAIK nothing on our code calls this endpoint, except the EOS server that uses it to verify the validity of the token we send. But, since we still use krb in production, I don't expect anything to break, but...

I've made the same change in puppet qa, needs to be merged to prod and jupyterhubs restarted. But since this should not affect anyone, I leave that for your consideration.

[OTG0076533](https://cern.service-now.com/service-portal?id=outage&n=OTG0076533)

### DIFF
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -165,6 +165,7 @@ jupyterhub:
           - profile
           - email
           - offline_access
+          - openid
         exchange_tokens: []
         logout_redirect_url: https://cern.ch/swan
         auto_login: True


etejedor:
Thanks for looking into this @diocas . I understand we still should be able to get cernbox tokens after the keycloak update you mentioned even without this change?

diocas:
All tokens can be requested as before. But access to that endpoint (which should not be used by anything besides EOS), will not work.

etejedor:
Ok, but then obtaining the EOS token will not work without this change iiuc -- which is not critical because we still rely on krb for EOS. But still, we'll get the errors in production.

We can group this together with the other changes that are in the pipeline (e.g. LCG 103), but those still require testing so perhaps May 15th is early. What do you think @diocas @dreyercito ?

diocas:
Getting the tokens should work, what fails is using them. But we don't use them.

etejedor (approved):

etejedor:
I merged, tagged the charts and deployed in QA. I do not observe anything wrong in the hub logs regarding the token management. @diocas you'll make a PR for puppet too?

PR #174, opened 2023-08-28
Use fluentd official helm chart

PMax5:
Separated the fluentd pipeline in three different config maps: sources, filters and outputs. The SWAN logs are collected from the source, filtered with custom filters and pushed to an endpoint defined in the outputs config map.

Furthermore, included a CA bundle file, which is mounted in the fluentd pods and  includes the CERN CA certificates, in order to be able to push the filtered logs to a custom endpoint through TLS and to be able to verify the TLS certificate of the server.


swan-cern/templates/fluentd/fluentd_outputs.conf.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: {{ .Release.namespace }}
+data:
+  outputs.conf: |-
+      # Push logs to MONIT endpoint
+      <match {logs,metrics}>
+          @type http
+          endpoint_url    {{ .Values.fluentd.output.endpoint }}
+          serializer      json
+          http_method     post
+          ssl_no_verify   false
```
etejedor, line 15: This is already false by default as per:

https://github.com/fluent-plugins-nursery/fluent-plugin-out-http

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: fluentd-sources-conf
+  namespace: {{ .Release.namespace }}
+data:
+  sources.conf: |-
+      # Do not collect fluentd logs to avoid infinite loops.
+      <match kubernetes.var.log.containers.**fluentd**>
+          @type null
+      </match>
+
+    {{ if not .Values.fluentd.output.includeInternal }}
```
etejedor, line 14: iirc this was present in the old fluentd chart, and we are now copying it here? And this is to discard the logs of all the kube-system namespace containers?

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -0,0 +1,42 @@
...
+      <match kubernetes.var.log.containers.**_kube-system_**>
+          @type null
+      </match>
+    {{ end }}  
+
+      # Read from container logs
+      <source>
```
etejedor, line 21: This source is to specify our main source of logs, and we need to do this here now because before it was done by the old fluentd chart?

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
   image: gitlab-registry.cern.ch/swan/docker-images/hadoop-token-generator:v2.1.2
 
 fluentd:
-  containerRuntime: containerd
+  enabled: true
```
etejedor, line 261: Is this necessary? Isn't it enabled by default? I can't find this attribute [here](https://github.com/fluent/helm-charts/blob/main/charts/fluentd/values.yaml).

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
   plugins:
     - fluent-plugin-rewrite-tag-filter
     - fluent-plugin-out-http
     - fluent-plugin-grok-parser
+    - fluent-plugin-route
+  tolerations:
+  - key: null
+    operator: Exists
+    effect: "NoSchedule"
```
etejedor, line 270: iiuc this is to allow the deployment of the fluentd daemonset on the master node. I wonder if we strictly need this. Currently, as this PR is now, we are discarding everything that comes from the kube-system namespace -- the master node only contains control plane pods of this namespace. Only if we decided to push logs from kube-system it would make sense to have this setting.

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+  readinessProbe:
+    initialDelaySeconds: 500
+  initContainers:
+  - name: disableipv6
+    image: ubuntu:20.04
+    command: ["/bin/bash", "-c", "sysctl -w net.ipv6.conf.all.disable_ipv6=1"]
+    securityContext:
+      privileged: true
+  volumeMounts:
+    - name: ca-certificate
```
etejedor, line 308: This list item should be indented two spaces to the left (see e.g. the init container above).

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+    - fluentd-outputs-conf
+  fileConfigs:
+    01_sources.conf: ""
+    02_filters.conf: ""
+    03_dispatch.conf: ""
+    04_outputs.conf: ""
+  livenessProbe:
+    initialDelaySeconds: 500
+  readinessProbe:
+    initialDelaySeconds: 500
```
etejedor, line 300: I would add a comment here to say that this is to allow ample time for the plugins to be installed when the pod is starting.

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+  configMapConfigs:
+    - fluentd-prometheus-conf
+    - fluentd-sources-conf
+    - fluentd-filters-conf
+    - fluentd-outputs-conf
+  fileConfigs:
+    01_sources.conf: ""
+    02_filters.conf: ""
+    03_dispatch.conf: ""
+    04_outputs.conf: ""
```
etejedor, line 296: Do we also need these to an empty string, to prevent fluentd from taking the settings of the upstream chart? Isn't it enough to define configMapConfigs?

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+  ## Prometheus Monitoring
+  metrics:
+    serviceMonitor:
+      enabled: true
+      additionalLabels:
+        release: prometheus-operator
+  resources:
+    requests:
+      cpu: 5m
+      memory: 700Mi
```
etejedor, line 286: What happens if we don't set these? Do we already get reasonable defaults?

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
-    </label>
+    cacert: /etc/ssl/ca-bundle.crt
+  podSecurityPolicy:
+    enabled: false
+  ## Prometheus Monitoring
+  metrics:
+    serviceMonitor:
+      enabled: true
+      additionalLabels:
+        release: prometheus-operator
```
etejedor, line 282: What is this for? Do the fluentd pods emit metrics?

If we decide that we need `enabled: true`, do we need the `additionalLabels` setting too (seems to be the default)?

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+  livenessProbe:
+    initialDelaySeconds: 500
+  readinessProbe:
+    initialDelaySeconds: 500
+  initContainers:
+  - name: disableipv6
+    image: ubuntu:20.04
+    command: ["/bin/bash", "-c", "sysctl -w net.ipv6.conf.all.disable_ipv6=1"]
+    securityContext:
+      privileged: true
```
etejedor, line 306: Is this taken from the CERN magnum configuration for fluentd? I guess it is needed?

swan-cern/templates/fluentd/fluentd_filters.conf.yaml
```diff
@@ -0,0 +1,131 @@
...
+      # Emit metric log records with type "metrics"
+      # Accessible in OpenSearch with pattern monit_private_swan_logs_metrics
+      <filter {metrics}>
+          @type record_transformer
+          <record>
+              type "metrics"
+          </record>
+      </filter>  
+
+      <filter {logs,metrics}>
```
etejedor, line 121: This filter block lost its original comment, see [here](https://github.com/swan-cern/swan-charts/blob/master/swan-cern/values.yaml#L382-L383).

etejedor (changes requested):
Many thanks for this work!

I left some comments to address before merging.

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+  volumeMounts:
+    - name: ca-certificate
+      mountPath: /etc/ssl/ca-bundle.crt
+      readOnly: true
+  volumes:
+    - name: ca-certificate
+      hostPath:
+        path: /etc/pki/tls/certs/ca-bundle.crt
```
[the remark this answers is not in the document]
diocas: Which will not be the case if you deploy it, say, in Oracle Cloud? Shouldn't we pass the ca when installing and stop replying on things of the host?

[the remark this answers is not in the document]
etejedor: That's a fair point. The advantage of getting it from the host is that we don't have to update it ourselves. @PMax5 can the TLS CA cert option of fluentd be a URL?

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -0,0 +1,42 @@
...
+      # Do not collect fluentd logs to avoid infinite loops.
+      <match kubernetes.var.log.containers.**fluentd**>
+          @type null
+      </match>
+
+    {{ if not .Values.fluentd.output.includeInternal }}
+      <match kubernetes.var.log.containers.**_kube-system_**>
+          @type null
+      </match>
+    {{ end }}
```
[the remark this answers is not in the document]
PMax5: Yes, that was also fixed.

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -0,0 +1,42 @@
...
+              keep_time_key false
+          </parse>
+          emit_unmatched_lines true
+      </source>
+    
+        # Add some basic metadata to the records
+      <filter kubernetes.**>
+        @type kubernetes_metadata
+        @id filter_kube_metadata
+      </filter>
```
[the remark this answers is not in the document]
PMax5: This is to add the kubernetes metadata (such as pod and host names) to the logs data.

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
PMax5: Yes, it was present in the old upstream chart and was copied here since it is not present in the official one. It is to discard the logs of all the kube-system namespace containers.

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
PMax5: Yes, that is correct. The official upstream chart does not have this defined by default.

swan-cern/values.yaml
PMax5: It was being used to not include the fluentd config maps (with the fluentd configuration) if we ever decided to disable it. After further review, the option was removed, since fluentd is part of the SWAN cern infrastructure and shouldn't be disabled.

swan-cern/values.yaml
PMax5: After reviewing it, it is not needed due to what you said - we are discarding everything that comes from the kube-system namespace already.

swan-cern/values.yaml
PMax5: This was fixed.

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
+    image: ubuntu:20.04
+    command: ["/bin/bash", "-c", "sysctl -w net.ipv6.conf.all.disable_ipv6=1"]
+    securityContext:
+      privileged: true
+  volumeMounts:
+    - name: ca-certificate
+      mountPath: /etc/ssl/ca-bundle.crt
+      readOnly: true
+  volumes:
+    - name: ca-certificate
```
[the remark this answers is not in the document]
PMax5: This was fixed.

swan-cern/values.yaml
PMax5: After review, these options are not needed. The readiness probe fails until all fluentd plugins are installed and is only marked as ready when all of them are installed and fluentd is initialized.

swan-cern/values.yaml
PMax5: Unfortunately, we need these to override the default fluentd configuration that is present in the upstream chart.

swan-cern/values.yaml
PMax5: I tested fluentd without those settings and it still works fine. Therefore, they were removed.

swan-cern/values.yaml
PMax5: Initially I thought this was related to the liveness and readiness probes but after further review, this is not needed and was removed.

swan-cern/values.yaml
```diff
@@ -258,154 +258,61 @@ hadoopTokenGenerator:
...
-
-    # Forward errors in this pipeline to fluentd pod stdout for debugging
-    <label @ERROR>
-        <match **>
-            @type stdout
-        </match>
-    </label>
+    cacert: /etc/ssl/ca-bundle.crt
+  podSecurityPolicy:
+    enabled: false
```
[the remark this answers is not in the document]
PMax5: After further review this was also removed, as it is not needed.

swan-cern/templates/fluentd/fluentd_filters.conf.yaml
PMax5: I added it back. Initially I had not noticed it had been lost.

swan-cern/values.yaml
etejedor: Ok, but let's remember this setting. In the past we had network problems in our cluster due to ipv6, I understand the magnum colleagues did this for a reason (which I don't know if it's still valid, we'll see).

etejedor: @diocas what you do think?

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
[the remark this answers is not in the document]
etejedor: Ok thanks, can we have the comment aligned with its filter block?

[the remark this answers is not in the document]
etejedor: Also, please say that in the comment: `Add some basic kubernetes metadata (e.g. pod name) to the logs`

swan-cern/files/ca-certs/ca-bundle.crt
```diff
@@ -0,0 +1,96 @@
...
+/DZhgGHRRoE4/D0kQMP2XTz6cClbNklECTlp0qZIbiaf350HbcDEFzYRSSIi0emv
+kRGcMgsi8yTTU87q8Cr4hETxAF3ZbSVNC0ZaTZ8RBbM9BXguhYzKkVBgG/cMpUjs
+B6tY2HMZbAZ3TKQRb/bRyUigM9DniKWeXkeL/0Nsno+XbcpAqLjtVIRwCg6jTLUi
+1NRsl3BP6C824dVaoI8Ry7m+o6O+mtocw4BMhHfTcoWCO8CWjT0ME67JzaAYa5eM
++OqoWtgbgweBlfO0/3GMnVGMAmI4FlhH2oWKWQgWdgr0Wgh9K05VcxSpJ87/zjhb
+MQn/bEojWmp6eUppPaqNFcELvud41qoe6hLsOYQVUQ1sHi7n6ouhg4BAbwS2iyD2
+uiA6FHTCeLreFGUzs5osPKiz3GE5D6V9she9xIQ=
+-----END CERTIFICATE-----
+
+# CERN Grid Certification Authority
```
etejedor, line 41: When does this certificate expire?

swan-cern/files/ca-certs/ca-bundle.crt
```diff
@@ -0,0 +1,96 @@
+# CERN Root Certification Authority 2
+-----BEGIN CERTIFICATE-----
```
etejedor, line 2: This expires in 2033 ?

swan-cern/values.yaml
```diff
@@ -258,154 +258,39 @@ hadoopTokenGenerator:
...
+    - fluentd-filters-conf
+    - fluentd-outputs-conf
+  fileConfigs:
+    # This is to disable the configuration that comes from upstream
+    01_sources.conf: "" 
+    02_filters.conf: ""
+    03_dispatch.conf: ""
+    04_outputs.conf: ""
+  volumeMounts:
+  # Installed by default on host VM
```
etejedor, line 282: This comment is no longer valid.

swan-cern/files/ca-certs/ca-bundle.crt
PMax5: Yes, the expiration time is: Mar 19 13:05:34 2033 GMT

swan-cern/values.yaml
```diff
@@ -258,154 +258,38 @@ hadoopTokenGenerator:
...
-        http_method     post
-    </match>
-
-    # Forward errors in this pipeline to fluentd pod stdout for debugging
-    <label @ERROR>
-        <match **>
-            @type stdout
-        </match>
-    </label>
+    cacert: /etc/ssl/ca-bundle.crt
```
etejedor, line 269: This path repeats also in the volume mount below, it would be nice to use an anchor to avoid this repetition:

https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors

swan-cern/values.yaml
PMax5: I forgot to remove it after I started to use the new certificate bundle. It was removed now.

swan-cern/templates/fluentd/fluentd_sources.conf.yaml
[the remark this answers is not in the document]
PMax5: I have changed the comment

swan-cern/values.yaml
PMax5: I added an achor so that the path is only defined in one place.

etejedor (approved):
Nice piece of work! Congrats!

PR #176, opened 2023-08-30
Hpc integration

diocas:
This PR follows #167, so only the last commit is relevant. Once the other is merged, I'll rebase.

It adds the possibility to mount a CephFS mount, which is necessary to add support for the HPC service (it's a shared space between submission and nodes).
It requires extra configuration, I'll add it in the ops repo later. This feature is disabled by default.


swan-cern/files/swan_hpc_config.py
```diff
@@ -0,0 +1,73 @@
+from kubernetes_asyncio.client.models import (
```
etejedor, line 1: After the last restructuring to unite condor with spark, shouldn't this go in [here](https://github.com/swan-cern/swan-charts/blob/master/swan-cern/files/swan_computing_config.py)?

diocas: Done, can you review and eventually merge? I need your help to config the gitops repo.
(This also needs some testing, as I didn't test the move to swan_computing yet...)

PR #186, opened 2023-10-23
Adapt to new version of the zero to jupyterhub chart (3.1.0)

etejedor:
This includes:
- Changes in the swan chart definition to adapt to the new version of the upstream chart
- Changes in the spawner-related hooks to be compatible with the async KubeSpawner

This goes together with https://github.com/swan-cern/jupyterhub-extensions/pull/82


swan-cern/values.yaml
```diff
@@ -148,10 +148,12 @@ swan:
           client_id: # placeholder, check secrets
           client_secret: # placeholder, check secrets
           oauth_callback_url: # placeholder, check secrets
-
           # skip refreshing tokens if already refreshed in last 110 minutes
           # this assumes tokens provided by keycloak are valid for 120 minutes
           auth_refresh_age: 6600
+          # Force old behaviour. This call is expensive and users most probably
```
etejedor, line 154: What is "old" here? Specify JH versions please (the one when it was not done, and the one we are updating to where we saw this new behaviour). I would put this info in the commit message, not in the yaml.

diocas: It was in both places. But removed it from the chart.

swan-cern/files/swan_config_cern.py
```diff
@@ -39,34 +56,29 @@ def _init_eos_secret(self):
...
+        }
+        secret_data.metadata = secret_meta
+        secret_data.data = {}
+        secret_data.data['krb5cc'] = eos_token_base64
 
+        try:
+            # eos-tokens secret is cleaned when user session ends, so try creating it
+            await self.spawner.api.create_namespaced_secret(swan_container_namespace, secret_data)
+        except ApiException:
+            # A secret with the same name exists, probably a remnant of a wrongly-terminated session, then replace it
```
diocas, line 77: Do we have a log object? This should be logged as a warning (since even wrongly terminated sessions should still remove it, no?)

swan-cern/files/swan_config_cern.py
```diff
@@ -39,34 +56,29 @@ def _init_eos_secret(self):
...
-            secret_data.metadata = secret_meta
-            secret_data.data = {}
-            secret_data.data['krb5cc'] = eos_token_base64
+        secret_data = V1Secret()
+
+        secret_meta = V1ObjectMeta()
+        secret_meta.name = eos_secret_name
+        secret_meta.namespace = swan_container_namespace
+        secret_meta.labels = {
+            "swan_user": username
```
diocas, line 67: Could we align the label name to the same as the container?

swan-cern/files/swan_config_cern.py
```diff
@@ -39,34 +56,29 @@ def _init_eos_secret(self):
...
+        }
+        secret_data.metadata = secret_meta
+        secret_data.data = {}
+        secret_data.data['krb5cc'] = eos_token_base64
 
+        try:
+            # eos-tokens secret is cleaned when user session ends, so try creating it
```
diocas, line 74: Btw, this reminds me that we split the code... This means that we have similar logic in 2 different places (we assume here that it will be created, but it's done somewhere else... And in that place we assumed it was created here...). I'm ok with leaving it here for now, but it's not the ideal solution. Let's open a PR upstream to fix the async call and then we put them back together (unless you think they should all live in the spawner).

swan/values.yaml
```diff
@@ -118,7 +118,8 @@ jupyterhub:
     # placeholder for hub secret token
     secretToken:
   hub:
-    fsGid: 0
+    podSecurityContext:
+      fsGroup: 0
```
diocas, line 122: Did you end up trying to deploy without this? I don't remember what was the conclusion of our discussion (if to leave it or not). But let's document it here in this PR, at least.

diocas (approved):
Some small comments, but nothing that I would say prevents the merger.

swan-cern/files/swan_config_cern.py
etejedor: Yes, creation and deletion are in different places now. I'd prefer they are all in the spawner, but that does not mean they can't be hooks. Regarding the contribution upstream, I added a comment to this task https://its.cern.ch/jira/browse/SWAN-85

swan/values.yaml
etejedor: No I did not try without in the end. I added this task to do it in the future: https://its.cern.ch/jira/browse/SWAN-92

swan-cern/files/swan_config_cern.py
etejedor: Wrongly terminated sessions should remove it, yes, since the removal code that is part of `stop` is in a `finally` block. I guess it could happen that the hub crashes and the `stop` method is never called for the user session. I can add the warning anyway, though, we should have access to the logger of the spawner from here.

swan-cern/files/swan_config_cern.py
etejedor: What name exactly?

PR #191, opened 2023-11-16
Turn Spark pod hook into a generic hook for handling external computing integrations

etejedor:
This is to implement a special treatment for HTCondor pools at the level of the modifier pod hook.

Now Spark and HTCondor will share a hook, which will configure both. The reason for this is that both share the need to open ports in the user session. Depending on what the user selects in the form (Spark, HTCondor or both) we will create a k8s service that provides the corresponding ports.

### DIFF
--- a/swan-cern/files/swan_computing_config.py
+++ b/swan-cern/files/swan_computing_config.py
@@ -63,7 +63,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
             return None
 
         username = self.spawner.user.name
-        hadoop_secret_name ='hadoop-tokens-%s' % username
+        hadoop_secret_name = f'hadoop-tokens-{username}'
 
         webhdfs_token_base64 = ''
         k8suser_config_base64 = ''
@@ -120,7 +120,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
             else:
                 await self.spawner.api.create_namespaced_secret(swan_container_namespace, secret_data)
         except ApiException as e:
-            raise Exception("Could not create required hadoop secret: %s\n" % e)
+            raise Exception(f'Could not create required hadoop secret: {e}\n')
 
         return hadoop_secret_name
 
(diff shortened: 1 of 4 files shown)


swan-cern/files/swan_computing_config.py
```diff
@@ -267,47 +286,59 @@ async def _init_spark(self, pod_labels):
...
+        return condor_pool != 'none'
 
-            # Define some 18 random NodePorts on the cluster for spark using V1Service
+    async def _open_ports(self, num_ports):
+        """
+        Create a service that opens the necessary ports for the computing integrations
+        (Spark, HTCondor)
+        """
+
+        computing_ports_service = "computing-ports" + "-" + self.spawner.user.name
```
diocas, line 302: Avoidable string concatenation. But I also prefer string interpolation.

swan-cern/files/swan_computing_config.py
```diff
@@ -267,47 +286,59 @@ async def _init_spark(self, pod_labels):
...
-                await self.spawner.api.delete_namespaced_service(spark_ports_service, swan_container_namespace)
-                service = await self.spawner.api.read_namespaced_service(spark_ports_service, swan_container_namespace)
+                service = await self.spawner.api.read_namespaced_service(computing_ports_service, swan_container_namespace)
             except ApiException:
                 # not existing, create
                 try:
                     service = await self.spawner.api.create_namespaced_service(swan_container_namespace, service_template)
                 except ApiException as e:
-                    raise Exception("Could not create service that allocates random ports for Spark in k8s cluster: %s\n" % e)
+                    raise Exception("Could not create service that allocates random ports for computing (Spark, HTCondor) integrations: %s\n" % e)
```
diocas, line 337: if we use "computing resources" we don't need to write their names which, in the future, might be outdated.

diocas (changes requested):

swan-cern/files/swan_computing_config.py
diocas: Would also interpolate for consistency.

etejedor: Agreed, it can get outdated easily. I will keep it generic.

etejedor: As for string interpolation, you want that I change this instance or everything in the file?

swan-cern/files/swan_computing_config.py
etejedor: Ok!

swan-cern/files/swan_computing_config.py
etejedor: Done in all the document.

diocas (approved):

PR #192, opened 2023-11-27
Configure new Alma9 image instead of CentOS7 via chart settings

etejedor:
With the newly added support for Alma9, two images will coexist (Alma9 and the old CentOS7). Here we configure Alma9 via chart settings and the URL of the CentOS7 image is provided via an option of our kubespawner.

Note: name and tag of the Alma9 image are subject to change before merging this PR.

### DIFF
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -81,8 +81,8 @@ jupyterhub:
     storage:
       type: none
     image:
-      name: "gitlab-registry.cern.ch/swan/docker-images/systemuser"
-      tag: "v6.0.0"
+      name: "gitlab-registry.cern.ch/swan/docker-images/jupyter/swan-cern"
+      tag: "v0.0.1"
       pullPolicy: "Always"
     cloudMetadata:
       # until we configure networkPolicy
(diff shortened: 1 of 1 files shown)


swan/values.yaml
```diff
@@ -81,8 +81,8 @@ jupyterhub:
     storage:
       type: none
     image:
-      name: "gitlab-registry.cern.ch/swan/docker-images/systemuser"
-      tag: "v6.0.0"
+      name: "gitlab-registry.cern.ch/swan/docker-images/swan-cern"
```
diocas, line 84: `swan-images/jupyter/swan-cern`

diocas (changes requested):

swan/values.yaml
etejedor: `gitlab-registry.cern.ch/swan/docker-images/jupyter/swan-cern` and tag `v0.0.1`?

diocas: 👍

diocas (approved):

PR #210, opened 2024-03-18
Change EOS chart used by SWAN

PMax5:
Use eosxd-csi chart for EOS as the old chart was no longer maintained (fusex). Furthermore, removed unused eos chart that was present in the SWAN chart Chart.yaml file.

This new chart allows to specify custom mountpoints on the nodes of the kubernetes cluster, which was a limitation of the fusex chart, meaning that it is possible to have two different deployments of EOS running in the same kubernetes cluster.


diocas (changes requested):
I'm not sure we should remove the fusex chart. Remember that the SWAN chart can be used outside of CERN, in installations that do not use the EOS instances we have here.
The new csi should allow us to configure it in the same way, but we do not have any documentation on how to do it.

etejedor (approved):

PR #216, opened 2024-04-22
Implement custom allocation for SWAN event participants with a GPU

etejedor:
Context: SWAN hosts events (i.e. trainings) that often require the access to GPUs from participants. Extra resources are provisioned to be able to support such events.

The functionality implemented by this commit allows to reserve some GPU resources for exclusive use of the participants of an event. Only pods from those participants (who must belong to an egroup) will be allocated on the event resources. This is useful to guarantee that the participants will have the resources that were agreed with the organisers.

Furthermore, if the GPU resources are fragments of MIG GPUs, now we can configure that event pods must request the desired type of fragment, so the matching is properly done at the GPU resource level too.


PMax5 (dismissed):

swan-cern/files/swan_computing_config.py
```diff
@@ -191,6 +271,12 @@ def _modify_containers_for_spark(self, hadoop_secret_name):
             ),
         )
 
+    def _gpu_enabled(self):
+        """
+        return True if the user has requested a GPU
+        """
+        return "cu" in self.spawner.user_options[self.spawner.lcg_rel_field]
```
diocas, line 278: Depending on how Rodrigo implements the custom env in the spawner, this might crash ^

swan-cern/files/swan_computing_config.py
```diff
@@ -393,4 +479,11 @@ def computing_modify_pod_hook(spawner, pod):
     return computing_pod_hook_handler.get_swan_user_pod()
 
 
+# Custom configuration options
+# Name of the role that is assigned to participants of events hosted by SWAN
+events_role = get_config('custom.events.role', 'swan-events')
```
diocas, line 484: Why did you declare the vars here instead of inside the method that uses it? It would be cleaner if in the future we remove that method.

diocas (commented):
Some last minute comments, in case you want to address them later.

swan-cern/files/swan_computing_config.py
etejedor: I followed the same structure as in the other hooks that have custom config. E.g. see https://github.com/swan-cern/swan-charts/blob/master/swan-cern/files/swan_config_cern.py#L232-L234

swan-cern/files/swan_computing_config.py
etejedor: If so, he'll fix it :) This is purely a code move.

PR #240, opened 2024-07-08
Prevention against KeyErrors

RodrigoSobral2000:
Since the code was initially projected to be dependent on LCG configurations, extra prevention measures need to be taken, since these LCG configurations are no longer present for custom environments.

### DIFF
--- a/swan-cern/files/swan_computing_config.py
+++ b/swan-cern/files/swan_computing_config.py
@@ -139,9 +139,6 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
 
         cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')
 
-        if cluster == 'none':
-            return None
-
         username = self.spawner.user.name
         hadoop_secret_name = f'hadoop-tokens-{username}'
 
@@ -314,6 +311,9 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
 
         cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
         max_mem = self.spawner.user_options[self.spawner.user_memory]
+
+        if cluster == 'none':
+            return
 
         # add basic spark envs
 
--- a/swan-cern/files/swan_computing_config.py
+++ b/swan-cern/files/swan_computing_config.py
@@ -137,9 +137,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
         Create secret for Spark/Hadoop
         """
 
-        cluster = 'none'
-        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
-            cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')
 
         if cluster == 'none':
             return None
@@ -288,9 +286,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
         """
 
         user_roles = self.spawner.user_roles
-        cluster = 'none'
-        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
-            cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')
 
         if cluster == "analytix" and "analytix" not in user_roles:
            raise ValueError(
@@ -316,10 +312,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
         notebook_container = self._get_pod_container('notebook')
         username = self.spawner.user.name
 
-        cluster = 'none'
-        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
-            cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
-
+        cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
         max_mem = self.spawner.user_options[self.spawner.user_memory]
 
         if cluster == 'none':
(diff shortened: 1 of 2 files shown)


swan-cern/files/swan_computing_config.py
```diff
@@ -137,7 +137,9 @@ async def _init_hadoop_secret(self):
         Create secret for Spark/Hadoop
         """
 
-        cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = 'none'
```
etejedor, line 140: This is usually done in a more pythonic way in a one-liner.

etejedor:
```suggestion
        cluster = self.spawner.user_options[self.spawner.spark_cluster_field] if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type else 'none'
```

etejedor: Or what if we just do `cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')`

RodrigoSobral2000: Isn't that worse for readability?

swan-cern/files/swan_computing_config.py
```diff
@@ -275,7 +277,7 @@ def _gpu_enabled(self):
         """
         return True if the user has requested a GPU
         """
-        return "cu" in self.spawner.user_options[self.spawner.lcg_rel_field]
+        return "cu" in self.spawner.user_options.get(self.spawner.lcg_rel_field, '')
```
etejedor, line 280: This is correct, but I'm thinking of the best way to do this. Didn't we say at some point that GPUs would become a first class citizen in the web form, meaning they would appear separately from memory and cpu, and therefore here we could just check if the form had a GPU selected? Instead of relying on the name of the LCG release to contain "cu".

swan-cern/files/swan_computing_config.py
```diff
@@ -286,7 +288,9 @@ def _spark_enabled(self):
         """
 
         user_roles = self.spawner.user_roles
-        cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = 'none'
+        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
```
etejedor, line 292: Here we can also just do `cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')`

swan-cern/files/swan_computing_config.py
```diff
@@ -312,7 +316,10 @@ async def _init_spark(self):
         notebook_container = self._get_pod_container('notebook')
         username = self.spawner.user.name
 
-        cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = 'none'
```
etejedor, line 319: Here actually we know that a spark cluster has been selected, otherwise this function would not even be executed (it is protected by `_spark_enabled()`, which already checks the field.

So we can just do `cluster = self.spawner.user_options[self.spawner.spark_cluster_field]`.

And below, we can remove:
```
        if cluster == 'none':
            return
```
I know this is not part of your changes, but I just realized we can optimize that part.

swan-cern/files/swan_computing_config.py
RodrigoSobral2000: yeah we meantioned that for the future, but that wouldn't be for now, right?

swan-cern/files/swan_computing_config.py
```diff
@@ -376,7 +383,9 @@ def _condor_enabled(self):
         """
         return True if the user has selected an HTCondor pool.
         """
-        condor_pool = self.spawner.user_options[self.spawner.condor_pool]
+        condor_pool = 'none'
+        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
```
etejedor, line 387:
```suggestion
        condor_pool = self.spawner.user_options.get(self.spawner.condor_pool], 'none')
```

swan-cern/files/swan_computing_config.py
RodrigoSobral2000: Cause that would involve changes in all repos, including charts, gitops, spawner, etc.
And we wouldnt want to conflict all those changes with the customenvs for now

swan/files/swan_config.py
```diff
@@ -24,9 +24,14 @@ def __init__(self, spawner, pod):
...
 
         # pod labels
-        pod_labels = dict(
-            lcg_release = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
-        )
+        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
+            pod_labels = dict(
+                lcg_release = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
+            )
+        else:
```
etejedor, line 31: else if self.spawner.user_options[self.spawner.software_source] ==  a custom env

swan/files/swan_config.py
```diff
@@ -24,9 +24,14 @@ def __init__(self, spawner, pod):
...
+            pod_labels = dict(
+                lcg_release = f'customenv-{self.spawner.builder}_{self.spawner.builder_version}'
+            )
```
etejedor, line 34: add else:
          pod_labels = {}

swan-cern/files/swan_computing_config.py
etejedor: The future is here: https://its.cern.ch/jira/browse/SWAN-215.

So yes, we leave it as it is here for now.

swan-cern/files/swan_computing_config.py
```diff
@@ -137,7 +137,7 @@ async def _init_hadoop_secret(self):
         Create secret for Spark/Hadoop
         """
 
-        cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
+        cluster = self.spawner.user_options.get(self.spawner.spark_cluster_field, 'none')
```
etejedor, line 140: I realize now that this change is not needed since `_init_hadoop_secret` is also protected by the `_spark_enabled` condition.

But what I would do is remove below:
```
 if cluster == 'none':
            return None
```

etejedor: Since the cluster can't be none here either, `_spark_enabled` already checked.

etejedor: As I said above, the change is not needed.

etejedor (approved):

PR #255, opened 2024-08-20
Use PVC instead of hostpath for EOS automounts

PMax5:


swan/files/swan_config.py
```diff
@@ -175,21 +175,20 @@ def swan_pod_hook(spawner, pod):
...
             name='eos',
             mount_path='/eos',
             mount_propagation='HostToContainer'
-        ),
-    )
-    c.SwanKubeSpawner.volumes.append(
-        V1Volume(
-            name='eos',
-            host_path=V1HostPathVolumeSource(
-                path=get_config("custom.eos.automountHostPath", "/var/eos")
```
etejedor, line 190: `custom.eos.automountHostPath` is not needed anymore, then? We can perhaps get rid of the anchor and just set it as a config argument for the eosxd-csi chart?

swan/templates/eos/automount-pvc.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
etejedor, line 1: We can call the file just `pvc.yaml`?

swan/files/swan_config.py
PMax5: I removed it yes

swan/templates/eos/automount-pvc.yaml
PMax5: That is also true, changed.

etejedor (approved):

PR #258, opened 2024-08-27
swan-cern: Add swan-cern-system chart for system components

PMax5:
This option ensures that chart resources are not deployed unless declared in the values. It was introduced following the goal to adopt a blue/green strategy when promoting new features in SWAN.

The resources in one of the environments will be allocated temporarily while a feature is being tested and will then be removed by toggling this chart flag.


etejedor (approved):
Looks good, but let's not merge for now. We have to wait until September when we will prepare for the switch to the new Kubernetes template.

etejedor (approved):

PR #279, opened 2024-09-26
Add network policies to restrict traffic in user pods

PMax5:
This change intends to increase security by only allowing the user pods to communicate with authorized resources, such as the hub and proxy and allow incoming traffic from specific origins, including spark and htcondor.

### DIFF
--- a/swan-cern/values.yaml
+++ b/swan-cern/values.yaml
@@ -58,6 +58,31 @@ swan:
       image:
         name: gitlab-registry.cern.ch/swan/docker-images/jupyter/swan-cern
         tag: v0.0.27
+      networkPolicy:
+        ingress:
+          # Allow user pods to receive incoming connections on the NodePort service port range.
+          # This supports use cases where servers run inside the user pod and need to receive
+          # connections (e.g. Spark, HTCondor via Dask).
+          - from:
+            - ipBlock:
+                cidr: 0.0.0.0/0
+                # Exclude connections from other pods in our cluster's pod network.
+                except:
+                  - 10.0.0.0/8
+                  - 172.16.0.0/12
+                  - 192.168.0.0/16
+            ports:
+            - protocol: TCP
+              port: 30000
+              endPort: 32767
+        egress:
+          # Allow connection from the user pod to the hadoop token generator to get Hadoop credentials.
+          - to:
+              - podSelector:
+                  matchLabels:
+                    app: hadoop-token-generator
+            ports:
+              - port: 80
     scheduling:
       userPods:
         nodeAffinity:
@@ -174,6 +199,16 @@ swan:
           description: Access to hub Prometheus metrics
           scopes: ['read:metrics']
           services: [prometheus-service-monitor]
+      networkPolicy:
+        ingress:
+          # Allow connection from prometheus to the hub, to scrape the metrics endpoint
+          - ports:
+              - protocol: TCP
+                port: 8081
+            from:
+              - podSelector:
+                  matchLabels:
+                    app.kubernetes.io/name: prometheus
     custom:
       cull:
         # 4 hours
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -55,11 +55,12 @@ jupyterhub:
       tag: "v0.0.14"
       pullPolicy: "Always"
     cloudMetadata:
-      # until we configure networkPolicy
-      blockWithIptables: true
       ip: 169.254.169.254
     networkPolicy:
-      enabled: false
+      enabled: true
+      egressAllowRules:
+        dnsPortsCloudMetadataServer: false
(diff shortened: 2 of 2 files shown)


swan-cern/values.yaml
```diff
@@ -58,10 +58,36 @@ swan:
...
+        enabled: true
+        ingress:
+          - from:
+            - ipBlock:
+                cidr: 0.0.0.0/0
+                except:
+                  - 10.0.0.0/8
+                  - 172.16.0.0/12
+                  - 192.168.0.0/16
+                  - 169.254.169.254/32
```
etejedor, line 71: Is this the metadata server? Why is this ingress?

Also, why the other three?

swan-cern/values.yaml
```diff
@@ -58,10 +58,36 @@ swan:
...
+            ports:
```
etejedor, line 72: This deserves a comment:
```
Allow user pods to receive incoming connections on the NodePort service port range.
This supports use cases where servers run inside the user pod and need to receive
connections (e.g. Spark, HTCondor via Dask).
```

swan-cern/values.yaml
```diff
@@ -58,10 +58,36 @@ swan:
...
+            - protocol: TCP
+              port: 30000
+              endPort: 32767
+        egress:
+          - to:
+              - podSelector:
```
etejedor, line 78:
```
Allow connection from the user pod to the hadoop token generator to get Hadoop credentials.
```

swan-cern/values.yaml
```diff
@@ -58,10 +58,39 @@ swan:
...
+      networkPolicy:
+        enabled: true
+        ingress:
+          # Allow user pods to receive incoming connections on the NodePort service port range.
+          # This supports use cases where servers run inside the user pod and need to receive
+          # connections (e.g. Spark, HTCondor via Dask).
+          - from:
+            - ipBlock:
+                cidr: 0.0.0.0/0
+                except:
```
etejedor, line 70: can we add a comment:
```
Exclude connections from other pods.
```

swan/values.yaml
```diff
@@ -174,7 +172,7 @@ jupyterhub:
           raise Exception("Please configure pre_spawn_hook")
         c.KeyCloakAuthenticator.pre_spawn_hook = pre_spawn_hook
     networkPolicy:
-      enabled: false
+      enabled: true
```
etejedor, line 175: So this is to enable network policies related to the hub?

swan-cern/values.yaml
```diff
@@ -58,10 +58,40 @@ swan:
...
+                  matchLabels:
+                    app: hadoop-token-generator
+            ports:
+              - port: 80
     scheduling:
       userPods:
         nodeAffinity:
           matchNodePurpose: ignore
+    proxy:
+      chp:
```
etejedor, line 92: chp?

swan-cern/values.yaml
```diff
@@ -174,6 +204,8 @@ swan:
           description: Access to hub Prometheus metrics
           scopes: ['read:metrics']
           services: [prometheus-service-monitor]
+      networkPolicy:
```
etejedor, line 207: How come we enable here again for the hub?

swan-cern/values.yaml
etejedor: Also, why do we enable the proxy netpols in swan-cern and not in swan?

etejedor: I would enable it for all components in the swan chart, and then in swan-cern I would add the part that is swan cern specific (Spark, HTCondor, hadoop token generator).

swan-cern/values.yaml
etejedor: Doesn't swan-cern inherit that setting if you set it in swan?

swan-cern/values.yaml
PMax5: Ah I see, I changed it now so CERN specific restrictions are in swan-cern and the general policies are in swan

swan-cern/values.yaml
PMax5: Yes, I set it now only in swan and the CERN specific policies in swan-cern

swan/values.yaml
```diff
@@ -55,11 +55,13 @@ jupyterhub:
...
-      # until we configure networkPolicy
-      blockWithIptables: true
       ip: 169.254.169.254
     networkPolicy:
-      enabled: false
+      enabled: true
+      egressAllowRules:
+        dnsPortsCloudMetadataServer: false
+        dnsPortsPrivateIPs: false
+
```
etejedor, line 64: Extra line that should be removed

PMax5: Removed

etejedor (approved):

PR #283, opened 2024-10-09
swan-cern: Update gpu-operator to version v24.6.2

PMax5:


swan-cern/values.yaml
```diff
@@ -268,29 +268,30 @@ fluentd:
...
+    version: v24.6.2
     defaultRuntime: containerd
     initContainer:
       repository: registry.cern.ch/kubernetes
       version: 11.4.2-base-ubi8
   driver:
     repository: registry.cern.ch/kubernetes
     image: nvidia-gpu-driver
-    version: "v550.54.15-6.6.13-200.fc39.x86_64"
+    version: "v550.54.15-6.8.4-200.fc39.x86_64"
```
etejedor, line 285: Can we have this as a separate commit, since it is not related to the gpu-operator update?

swan-cern/values.yaml
```diff
@@ -268,29 +268,30 @@ fluentd:
...
     imagePullPolicy: Always
     manager:
       repository: registry.cern.ch/kubernetes
       version: v0.6.0
     licensingConfig:
       configMapName: "nvidia-grid-license"
-    kernelModuleConfig:
-      name: kernel-module-params
+      nlsEnabled: false
+  nfd:
```
etejedor, line 293: These changes are also unrelated to the update to fix the security issue? And also others below.

PMax5: It is to keep our chart in sync with the cern-magnum one. Should they be removed?

etejedor: What we can do, not to complicate things too much, is to separate this in two commits: one that updates the chart up until before the security issue fix, and one that strictly has the changes to fix the security issue.

swan-cern/values.yaml
```diff
@@ -268,29 +268,30 @@ fluentd:
           path: ca-bundle.crt
 
 gpu-operator:
-  enabled: true
+  enabled: false
```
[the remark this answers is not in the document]
PMax5: Yes, I missed that but changed it now

swan-cern/values.yaml
PMax5: Yeah that sounds good

etejedor (approved):

PR #286, opened 2024-10-10
swan-cern: Remove node-feature-discovery values

PMax5:
The node-feature-discovery component values are no longer under the gpu-operator ones and are now defined seperately in the upstream cern-magnum chart. This commit reflects such change

### DIFF
--- a/swan-cern/values.yaml
+++ b/swan-cern/values.yaml
@@ -351,8 +351,3 @@ gpu-operator:
     config:
       name: nvidia-device-plugin-config
       default: "default"
-  node-feature-discovery:
-    image:
-      repository: registry.cern.ch/kubernetes/node-feature-discovery
-      tag: v0.10.1
-


swan-cern/values.yaml
```diff
@@ -351,8 +351,3 @@ gpu-operator:
     config:
       name: nvidia-device-plugin-config
       default: "default"
-  node-feature-discovery:
-    image:
```
etejedor, line 355: Do we still need to include this somewhere, though? In the scope of which chart does cern-magnum define it?

PMax5: In the scope of their own chart, here: https://gitlab.cern.ch/kubernetes/automation/releases/cern-magnum/-/blob/master/values.yaml?ref_type=heads#L608

etejedor: But this means we do not deploy anymore any component that needs these settings? There's no node-feature-discovery anymore?

PMax5: I believe it is deployed with the CERN magnum chart, at cluster creation time

etejedor: I see, and before it was deployed by the gpu operator chart? So after this update we won't have the node-feature-discovery pods anymore in our cluster until we move to the new template?

PMax5: It will still be deployed, but through cern-magnum, not the gpu-operator

etejedor: But the cern-magnum of our current cluster is the old cern-magnum which does not have this setting, right? That's why I'm asking if before the node-feature-discovery was deployed via the gpu operator chart.

PR #306, opened 2025-01-27
[swan,swan-cern] Allocate the GPU flavour selected by the user

etejedor:
This is the charts side of a series of changes that allow the user to select a GPU flavour in the web form before starting their session.

When receiving the GPU selection from the user, a call to the k8s API is done to retrieve all the pods that have allocated a GPU. With such information, combined with the information about available GPU flavours in the cluster, we can see if the requested GPU flavour is free. If not, the user gets a message with all the free flavours (if any).

These changes also include adding extra capabilities to the hub service account to be able to get information about nodes in the cluster. This is necessary to retrieve the available GPU flavours in the cluster, which are currently specified as node labels.

### DIFF
--- a/swan-cern/files/swan_computing_config.py
+++ b/swan-cern/files/swan_computing_config.py
@@ -40,7 +40,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
 
         if self._gpu_enabled():
             # Configure GPU allocation
-            self._modify_pod_for_gpu()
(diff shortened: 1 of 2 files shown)


PMax5 (approved):

diocas (approved):

etejedor:
Superseded by https://github.com/swan-cern/swan-charts/pull/380, closing.

PR #309, opened 2025-03-05
Custom rules for Falco

rodrigo-sobral:
These two rules are made for tracking users activity, including:
- Logging each executed command via jupyter terminal;
- Logging every incoming and outcoming connection established with the user's pod.

The change on grok rules is made for not allowing the filtering rules to ignore falco logs (because falco pods are deployed within the kube-system namespace, which is being currently ignored by the `ignoreInternal` value). So a new value `internalFalco` was created in order to open an exception to this generic rule.

### DIFF
--- a/swan-cern-system/templates/fluentd/fluentd_filters.conf.yaml
+++ b/swan-cern-system/templates/fluentd/fluentd_filters.conf.yaml
@@ -24,15 +24,37 @@ data:
           </rule>
           <rule>
               key $.kubernetes.pod_name
+              pattern /.*-falco-.*/
+              tag falco
+          </rule>
+          <rule>
+              key $.kubernetes.namespace_name
+              pattern /kube-system/
+              tag system
+          </rule>
+          <rule>
+              key $.kubernetes.pod_name
               pattern /.?/
               tag logs
           </rule>
       </match>
+ 
+      {{ if not .Values.fluentd.output.includeInternal }}
+        <match {system}>
+            @type null
+        </match>
+      {{ end }}
+
+      {{ if not .Values.fluentd.output.includeFalco }}
+        <match {falco}>
+            @type null
+        </match>
+      {{ end }}
 
       # Retag user and hub logs to:
       # - metrics: jupyter and jupyterhub custom log messages, used to emit metrics
       # - logs: rest of log messages
-      <match {user,hub}>
+      <match {user,hub,falco}>
           @type rewrite_tag_filter
           capitalize_regex_backreference true
           <rule>
--- a/swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
+++ b/swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
@@ -10,12 +10,6 @@ data:
           @type null
       </match>
 
-      {{ if not .Values.fluentd.output.includeInternal }}
-      <match kubernetes.var.log.containers.**_kube-system_**>
-          @type null
-      </match>
-      {{ end }}  
-
       # Read from container logs
       <source>
           @type tail
--- a/swan-cern/README.md
+++ b/swan-cern/README.md
@@ -203,6 +203,16 @@ rmmod nouveau
 kubectl create -f https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/1.0.0-beta4/nvidia-device-plugin.yml
 ```
 
+### Provide custom rules for Falco deployment
+
+Create a custom rules file `falco-custom-rules.yaml` and encode it in **base64**. Then, through Magum, provide the encoded file as a label to the Falco chart. This way, the custom rules will be loaded and used by Falco.
+
+For more information, read the [CERN Kubernetes Documentation](https://kubernetes.docs.cern.ch/docs/security/falco/#writing-your-own-alerts).
+
+```bash
+--labels cern_chart_user_values="$(cat falco-custom-rules.yaml | base64 -w0)" \
+```
+
 ### Demo of upstream JupyterHub (no SWAN, no EOS, no CVMFS)
 This option uses upstream [JupyterHub Helm Chart](https://jupyterhub.github.io/helm-chart/)
 
--- a/swan-cern/templates/falco-custom-rules.yaml
+++ b/swan-cern/templates/falco-custom-rules.yaml
@@ -1,0 +1,49 @@
+customRules:
+    user_pod_rules.yaml: |-
+      - list: ignored_networks
+        items: ['"127.0.0.1/8"']
+
+      - macro: connection_filter
+        condition: >
+          fd.l4proto = tcp and (fd.typechar = 4 or fd.typechar = 6) and
+          (fd.ip != "0.0.0.0" and not fd.snet in (ignored_networks))
+
+      - macro: outbound_connection_tcp
+        condition: evt.type = connect and connection_filter and (evt.rawres >= 0 or evt.res = EINPROGRESS)
+      - macro: inbound_connection_tcp
+        condition: evt.type = accept and connection_filter
+
+      - macro: user_pod
+        condition: >
+          container
+          and k8s.pod.name startswith "jupyter-"
+          and k8s.ns.name = swan
+
+      - rule: Contact K8S API Server From Container
+        enabled: false
+        override:
+          enabled: replace
+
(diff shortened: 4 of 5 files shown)


swan-cern/README.md
```diff
@@ -203,6 +203,12 @@ rmmod nouveau
 kubectl create -f https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/1.0.0-beta4/nvidia-device-plugin.yml
 ```
 
+Provide custom rules for Falco deployment
```
etejedor, line 206: Can you make this a section at the same level of e.g. Monitoring, this means with `###`.

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,37 @@
...
+      output: >
+        user: %user.name --- pid: %proc.pid --- exec: %proc.execpath --- args: %proc.args --- cwd: %proc.cwd
+      priority: NOTICE
+      tags: [user_activity, k8s, monitoring]
+
+    - rule: Detect Incoming Connections in Jupyter Pod
+      desc: Detects incoming connections to Jupyter pod
+      condition: >
+        user_pod and evt.type in (accept,listen) and
+        fd.sip exists and fd.cip exists and
```
etejedor, line 21: What do the conditions on this line mean?

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,37 @@
...
+        evt.res = SUCCESS
```
etejedor, line 22: Why is this necessary?

swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -12,9 +12,19 @@ data:
 
       {{ if not .Values.fluentd.output.includeInternal }}
       <match kubernetes.var.log.containers.**_kube-system_**>
+        {{ if not .Values.fluentd.output.includeFalco }}
```
etejedor, line 15: Reverse the if else to prevent the unnecessary negation?

swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -12,9 +12,19 @@ data:
...
           @type null
+        {{ else }}
+          <filter kubernetes.var.log.containers.**_kube-system_**>
```
etejedor, line 18: Maybe a comment would be necessary here? Something like keep the logs of the falco pod that outputs the audit messages.

swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
```diff
@@ -12,9 +12,19 @@ data:
...
+            @type grep
+            <regexp>
+              key $.kubernetes.pod_name
```
etejedor, line 21: So this system makes sure that ONLY the logs of this Falco pod go through, and the rest of the logs of all the remaining pods in kube-system are discarded?

swan-cern/templates/falco-custom-rules.yaml
rodrigo-sobral: Without this condition, failed connection attempts would also be logged. By filtering only successful events, the rules focus on actual established network connections.

swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
etejedor: Still needs reversing.

swan-cern-system/templates/fluentd/fluentd_sources.conf.yaml
etejedor: Still missing comment.

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,37 @@
...
+        and k8s.pod.name startswith "jupyter-"
+        and k8s.ns.name = swan
+
+    - rule: Log User Commands in Jupyter Pod
+      desc: Logs all commands executed by the user
+      condition: user_pod and spawned_process
+      output: >
+        user: %user.name --- pid: %proc.pid --- exec: %proc.execpath --- args: %proc.args --- cwd: %proc.cwd
+      priority: NOTICE
+      tags: [user_activity, k8s, monitoring]
```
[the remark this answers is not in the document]
etejedor: Ok, I think you actually did not correct this PR yet, I'll wait then.

swan-cern-system/templates/fluentd/fluentd_filters.conf.yaml
```diff
@@ -22,17 +22,39 @@ data:
...
+          <rule>
+              key $.kubernetes.pod_name
+              pattern /.*-falco-.*/
+              tag falco
+          </rule>
+          <rule>
+              key $.kubernetes.namespace_name
+              pattern /kube-system/
+              tag system
+          </rule>
```
etejedor, line 34: According to these two rules, Falco logs will be tagged both `falco` and `system`. Doesn't this mean that below they will be match as `system` and therefore ignored (line 44)?

rodrigo-sobral: According to what I found on the internet, a log entry cannot have multiple tags simultaneously—it gets retagged based on the first matching rule. So, considering the tag `falco` is applied first, so the `system` rule does not apply.

etejedor: Can you double check this? What I have found is that a log line matches two rules and is tagged by both, only the last tag remains.

etejedor: It should be easy to test. If the current setup allows you to have Falco log lines on Monit, searchable through OpenSearch, but not the logs of any other component in kube-system, then it's ok.

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+            (evt.rawres >= 0 or evt.res = EINPROGRESS)
+          and not proc.name in (sshd, systemd, falco))
+
+      - macro: inbound_connection_tcp
+        condition: >
+          ( (evt.type = accept and evt.dir=> and fd.l4proto = tcp) and
+            (fd.typechar = 4 or fd.typechar = 6) and
+            (fd.ip != "0.0.0.0" and not fd.snet in (ignored_networks)) and
+            (evt.rawres >= 0 or evt.res = EINPROGRESS)
+          and not proc.name in (sshd, systemd, falco))
```
etejedor, line 26: To check what proc.name is here

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - rule: Contact K8S API Server From Container
+        enabled: false
+        override:
+          enabled: replace
+
+      - rule: Log User Commands in Jupyter Pod
+        desc: Logs all commands executed by the user
+        condition: user_pod and spawned_process
+        output: >
+          pod_name: %k8s.pod.name --- pname: %proc.pname --- cmdline: %proc.cmdline (%container.info)
```
etejedor, line 37: Better to know the full path of the binary being executed.

etejedor: Exec path, args, proc.pname, container_info

swan-cern/templates/falco-custom-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+        output: >
+          Incoming connection [%proc.pname]: %fd.rip:%fd.rport -> %fd.lip:%fd.lport (%container.info)
+        priority: NOTICE
+        tags: [user_activity, network, incoming]
+
+      - rule: Detect Outgoing Connections in Jupyter Pod
+        desc: Detects when a process initiates an outbound network connection
+        condition: user_pod and outbound_connection_tcp
+        output: >
+          Outgoing connection [%proc.pname]: %fd.lip:%fd.lport -> %fd.rip:%fd.rport (%container.info)
```
etejedor, line 53: First process name, then parent name

etejedor (approved):

PR #322, opened 2025-04-25
Remove offline_access scope from KeyCloakAuthenticator

laurall974:
Allows to prevent silent re-login after the refresh token expired.

### DIFF
--- a/swan/values.yaml
+++ b/swan/values.yaml
@@ -146,7 +146,6 @@ jupyterhub:
         scope:
           - profile
           - email
-          - offline_access
           - openid
         exchange_tokens: []
         auto_login: True


swan/values.yaml
```diff
@@ -146,7 +146,6 @@ jupyterhub:
         scope:
           - profile
           - email
-          - offline_access
```
etejedor, line 149: Can you explain in the commit message why we remove this scope? I.e. for the refresh token to have an expiration time of 12h, so that we force reauthentication after that time according to the new security rules in IT.

laurall974: Done

etejedor (approved):

PR #323, opened 2025-04-28
swan: New value for defining the domain name of the ATS hub instance

rodrigo-sobral:


etejedor:
Not needed as per https://github.com/swan-cern/jupyterhub-extensions/pull/140/commits/d010ab391951b1603178f4116f98039bf3140376#r2063302379

PR #326, opened 2025-05-05
swan-cern-system: New includeFalco value

rodrigo-sobral:
This change is made for not allowing the filtering rules to ignore falco logs (because falco pods are deployed within the kube-system namespace, which is being currently ignored by the `ignoreInternal` value). So a new value `internalFalco` was created in order to open an exception to this generic rule.

### DIFF
--- a/swan-cern-system/templates/fluentd/fluentd_filters.conf.yaml
+++ b/swan-cern-system/templates/fluentd/fluentd_filters.conf.yaml
@@ -24,15 +24,37 @@ data:
           </rule>
           <rule>
               key $.kubernetes.pod_name
+              pattern /.*-falco-.*/
+              tag falco
+          </rule>
+          <rule>
+              key $.kubernetes.namespace_name
+              pattern /kube-system/
+              tag system
+          </rule>
+          <rule>
+              key $.kubernetes.pod_name
               pattern /.?/
               tag logs
           </rule>
       </match>
+ 
+      {{ if not .Values.fluentd.output.includeInternal }}
+        <match {system}>
+            @type null
+        </match>
+      {{ end }}
+
+      {{ if not .Values.fluentd.output.includeFalco }}
+        <match {falco}>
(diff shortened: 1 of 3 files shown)


etejedor:
Thanks @rodrigo-sobral , I see this as the very first thing we can merge in production, since even if it has undesired side effects, it would not affect the users.

@PMax5 do you think we could even merge this on Thursday as part of the GPU operator upgrade?

etejedor (approved):

PMax5:
> Thanks @rodrigo-sobral , I see this as the very first thing we can merge in production, since even if it has undesired side effects, it would not affect the users.
> 
> @PMax5 do you think we could even merge this on Thursday as part of the GPU operator upgrade?

Yeah I believe we can. However, I would suggest we set the `includeFalco` value to `false` in our GitOps values since we won't need this type of logs in our main production instance.

etejedor:
> Yeah I believe we can. However, I would suggest we set the includeFalco value to false in our GitOps values since we won't need this type of logs in our main production instance.

Yes this is a good point, but then perhaps we should set the default in the swan-cern-system chart to `false`, and only set it to true for swan-acc-qa?

PMax5:
> > Yeah I believe we can. However, I would suggest we set the includeFalco value to false in our GitOps values since we won't need this type of logs in our main production instance.
> 
> Yes this is a good point, but then perhaps we should set the default in the swan-cern-system chart to `false`, and only set it to true for swan-acc-qa?

Yeah that may be better indeed. This way we only ensure that it stays active for that specific instance.

etejedor:
> Yeah that may be better indeed. This way we only ensure that it stays active for that specific instance.

Ok thanks, @rodrigo-sobral could you make the change to this PR to have the default `includeFalco` as false, then create a PR for gitops to set it to true only for swan-acc-qa for now?

PMax5 (approved):

PR #327, opened 2025-05-06
Swan-on-TN

rodrigo-sobral:
This PR includes relevant changes regarding preventions against not passed arguments like `condor_pool` or `lcg_rel_field` for custom environments and not allowed users from accessing the Technical Network. Along with the configuration for the new `swan_accpy` instance, in order to fill all the requirements imposed by the Swan-on-TN project.


swan-cern/files/swan_computing_config.py
```diff
@@ -139,9 +139,6 @@ async def _init_hadoop_secret(self):
 
         cluster = self.spawner.user_options[self.spawner.spark_cluster_field]
 
-        if cluster == 'none':
```
etejedor, line 142: Can you remind me why we could remove this?

swan/values.yaml
```diff
@@ -166,6 +166,7 @@ jupyterhub:
         # set home directory to EOS
         local_home: False
         centos7_image: "gitlab-registry.cern.ch/swan/docker-images/systemuser:v6.0.0"
+        accpy_image: "gitlab-registry.cern.ch/swan/docker-images/jupyter/swan-acc:v0.0.1"
```
etejedor, line 169: Once you have done the changes I requested for its Dockerfile, you could bump this version to `v1.0.0`

swan/files/swan_config.py
```diff
@@ -24,9 +24,16 @@ def __init__(self, spawner, pod):
...
-        pod_labels = dict(
-            lcg_release = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
-        )
+        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
+            pod_labels = dict(
+                lcg_release = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
+            )
+        elif self.spawner.user_options[self.spawner.software_source] == self.spawner.customenv_special_type:
+            pod_labels = dict(
+                lcg_release = f'customenv-{self.spawner.builder}_{self.spawner.builder_version}'
```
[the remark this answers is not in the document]
rodrigo-sobral: At the time this was changed, I was not sure if this label was being used as it is by another service. So I kept the name. I'll change it to `software_source`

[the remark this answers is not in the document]
etejedor: Before you change it, can you check whether the label `lcg_release` is used by us anywhere in our code?

swan-cern/files/swan_computing_config.py
rodrigo-sobral: If i'm not mistaken, at the time this was change, was to make just the custom environments work, because with didn't have this change https://github.com/swan-cern/jupyter-images/blob/swan-on-tn/swan-cern/scripts/before-notebook.d/10_spark.sh#L39

swan/files/swan_config.py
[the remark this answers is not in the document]
rodrigo-sobral: As far as I know, I checked all the repos I have, and none of them makes any use of the expression `lcg_release`

swan/files/swan_config.py
```diff
@@ -24,9 +24,16 @@ def __init__(self, spawner, pod):
     def get_swan_user_pod(self):
 
         # pod labels
-        pod_labels = dict(
-            lcg_release = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
-        )
+        if self.spawner.user_options[self.spawner.software_source] == self.spawner.lcg_special_type:
+            pod_labels = dict(
+                software_source = self.spawner.user_options[self.spawner.lcg_rel_field].split('/')[0]
```
etejedor, line 29: To be checked if the previous label here (lcg_release) was used somewhere else in our code.

etejedor: I think this label (lcg_release) was used for setting affinity with other pods that had the same LCG release some time ago:

https://github.com/swan-cern/swan-charts/commit/d8fb2dcdf9eab5e64cb5ccf78ea714d9facb9ecf

So it should not hurt to transform it into software_source. I can't find any occurrence of lcg_release at the moment anywhere else in the code.

etejedor (approved):

etejedor (approved):

PR #330, opened 2025-05-09
Declarative liqo peering for argocd

laurall974:
This charts create a set of CRs describing the peering with a remote cluster, which is automatically set up once the CRs are applied on both sides.
Docs: https://docs.liqo.io/en/v1.0.0/advanced/peering/peering-via-cr.html

### DIFF
--- a/liqo-peering/.helmignore
+++ b/liqo-peering/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/liqo-peering/Chart.yaml
+++ b/liqo-peering/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: liqo-peering
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.0.0"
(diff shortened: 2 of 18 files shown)


liqo-peering/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
etejedor, line 1: I believe the commit message needs an explanation of what you are adding here and why. It needs to explain what this new liqo-peering chart does and why we need it.

liqo-peering/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: liqo-peering
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
```
etejedor, line 5: I would remove all these comments that are standard for charts.

liqo-peering/templates/add-taint-label-to-virtualnode.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if eq .Values.role "consumer" }}
```
etejedor, line 1: It would be nice to add one/two lines of comment at the beginning of each yaml template to briefly explain what it does.

liqo-peering/values.yaml
```diff
@@ -0,0 +1,20 @@
+role: consumer  # or provider
+
+clusterId: local
+privateWireguardKey: # placeholder for private wireguard key
+publicWireguardKey: # placeholder for public wireguard key
+
+remoteCluster: 
+  clusterId: remote
+  userCommonName: "user@cluster"  # The CN of the user in the certificate
```
etejedor, line 9: I'd prefer not to have fake values in the chart base values.yaml, better to have it empty and force definition (in our case in gitops).

liqo-peering/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+  podCidr: # placeholder for pod CIDR
+  publicWireguardKey: # placeholder for public wireguard key
+  kubeconfig: |
+    # Here, paste your kubeconfig YAML content in base64 encoded!!!
+    apiVersion: v1
+    clusters:
+    ... ect
```
etejedor, line 16: Can we not have this as a fake value, but just add one line of comment? E.g. # kubeconfig, base64 encoded

swan-cern/templates/config.yaml
```diff
@@ -4,6 +4,8 @@ kind: ConfigMap
 metadata:
   name: swan-scripts-cern
   namespace: {{ .Release.Namespace }}
+  annotations:
+    liqo.io/allow-reflection: "true"
```
etejedor, line 8: Is this change related to peering or is it an annotation we need for this config map to be propagated?

swan/templates/config.yaml
```diff
@@ -3,6 +3,8 @@ kind: ConfigMap
 metadata:
   name: swan-scripts
   namespace: {{ .Release.Namespace }}
+  annotations:
```
etejedor, line 6: Same comment here.

swan-cern/templates/config.yaml
laurall974: It is an annotation we need for this configmap to be propagated!

laurall974: I put it in a separate commit to make things clearer!

swan/templates/config.yaml
laurall974: I put it in a separate commit to make things clearer!

swan-cern/templates/config.yaml
```diff
@@ -4,6 +4,8 @@ kind: ConfigMap
 metadata:
   name: swan-scripts-cern
   namespace: {{ .Release.Namespace }}
+  annotations:
```
etejedor, line 7: I guess having these annotations in the swan and swan-cern charts does not have any side effect when these CM objects are created and liqo is not even deployed?

laurall974: I assume that if Liqo is not present, the annotation is simply ignored by Kubernetes and it’s just treated as regular metadata.

PR #350, opened 2025-05-19
swan-cern-system: Add fluentd tag to differentiate logs

PMax5:
Add new community tag to differentiate and filter logs coming from different instances of SWAN, such as the general instance and the ATS one while using the production and qa monit producers


etejedor (approved):

etejedor:
Holding until we have tested a solution for the metrics that can work together with this option too -- for the logs we already know it can.

PR #380, opened 2025-06-17
[swan, swan-cern]: See available gpus flavour (rebase from Enric PR)

laurall974:
The goal is to show to the user only the GPU models that are free. 
- If the user try to create a session while no GPU are available -> an error will be show.
- "watch" permission has been added to pod in the SWAN namespace.

### DIFF
--- a/swan-cern/files/swan_computing_config.py
+++ b/swan-cern/files/swan_computing_config.py
@@ -40,7 +40,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
 
         if self._gpu_enabled():
             # Configure GPU allocation
-            self._modify_pod_for_gpu()
+            await self._modify_pod_for_gpu()
 
         if self._spark_enabled():
             # Configure Spark clusters at CERN
@@ -62,7 +62,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
 
         return self.pod
 
-    def _modify_pod_for_gpu(self):
+    async def _modify_pod_for_gpu(self):
         """
         Configure a pod that requested a GPU.
 
@@ -76,7 +76,8 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
         scheduled on resources that have been allocated for the event (and
         therefore have been labeled and tainted to host only event pods).
         """
-        if events_role in self.spawner.user_roles:
+        spawner = self.spawner
+        if spawner.SWAN_EVENTS_ROLE in spawner.user_roles:
             # The user is a participant of an event hosted by SWAN.
             # Their pod must be allocated on a node that has been
             # provisioned exclusively for the event
@@ -88,7 +89,7 @@ class SwanComputingPodHookHandler(SwanPodHookHandlerProd):
             # Add affinity to nodes that have been provisioned for the
             # event, i.e. labeled with the events role name
             node_selector_req = V1NodeSelectorRequirement(
-                key = events_role,
+                key = spawner.SWAN_EVENTS_ROLE,
                 operator = 'Exists'
             )
             node_selector_term = V1NodeSelectorTerm(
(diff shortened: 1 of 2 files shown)


swan-cern/files/swan_computing_config.py
```diff
@@ -38,6 +38,13 @@ async def get_swan_user_pod(self):
 
         required_ports = 0
 
+        if self._cuda_selected_without_gpu():
+            # Check if CUDA stack is selected but no GPU is provided
+            raise ValueError(
+                "CUDA software stack requires a GPU to be selected. "
+                "Please go back and select a GPU. If no GPUs are available, choose a different software stack."
+            )
+
```
etejedor, line 47: Instead of this, what I would do is modify the form in the frontend so that, when a CUDA stack is selected, a GPU is selected too--any, the first in the list. I might already do this as part of my changes? If not, please add.

swan-cern/files/swan_computing_config.py
```diff
@@ -114,30 +121,9 @@ async def _modify_pod_for_gpu(self):
...
-                    # ... but not the one requested by the user, inform them
-                    error_message = f'The selected GPU flavour ({gpu_description}) is not available. Please select one of the following:'
-                    error_message += '<ul>'
-                    for flavour in free_flavours:
-                        error_message += f'<li>{flavour}</li>'
-                    error_message += '</ul>'
-                    raise ValueError(error_message)
-            else:
-                # No GPUs available, inform the user
-                raise ValueError('Unfortunately, no GPUs are available at the moment. Please try again later.')
```
etejedor: If you remove this code, what happens in the case that two users request the same flavor at the same time, and only one GPU is left for that flavor? This will cause a race condition in which one will get the GPU, and the other will see a generic timeout error after 60 seconds, because the pod could not be scheduled. This might be good enough, I'm just bringing up this subject to discuss it.

swan-cern/files/swan_computing_config.py
laurall974: Yes, you're right, I originally added that line because if no GPUs were available, the field would show "None" and the user could still click on "Create Session." However, I’ve updated the logic: I reverted to your approach of checking the currently available GPUs directly (to avoid race conditions), and I’ve also decoupled _gpu_enabled from the selected GPU field. Now, it simply checks whether the selected software stack is CUDA, which prevent the behavior I mentioned before.

swan-cern/files/swan_computing_config.py
laurall974: I put this code back (cf previous comment), indeed I was not preventing this! One of the use would have get the normal generic timeout error.

swan-cern/files/swan_computing_config.py
```diff
@@ -172,6 +174,17 @@ async def _modify_pod_for_gpu(self):
             )
             self.pod.spec.tolerations = tolerations
 
+            # decrease GPU availability count
+            try:
+                with spawner.gpus.get_lock():
+                    if gpu_info and gpu_info.available > 0:
```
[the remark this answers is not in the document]
etejedor: And since this runs before the pod is actually created, here you DO need to decrement after you do the query to the k8s api server, IF for that particular flavor `free` is bigger than 0.

swan-cern/files/swan_computing_config.py
```diff
@@ -329,7 +342,8 @@ def _gpu_enabled(self):
         """
         return True if the user has requested a GPU
         """
-        return self.spawner.user_options[self.spawner.gpu] != 'none'
+        lcg_stack = self.spawner.user_options.get('lcg', '')
+        return 'cuda' in lcg_stack.lower()
```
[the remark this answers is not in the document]
laurall974: if the user create a session but no gpu are available the default value is "none" so I never go through the _modify_pod_for_gpu (which check that indeed pod are available or not)

etejedor (dismissed):

PMax5 (approved):

PR #402, opened 2025-08-05
swan-cern: filter the free gpu display in case of concurrent session request

laurall974:


swan-cern/files/swan_computing_config.py
```diff
@@ -127,8 +127,10 @@ async def _modify_pod_for_gpu(self):
                         spawner.log.info(f'Decreased currently free count for {gpu_description}: {gpu_info.free}/{gpu_info.count} available')
                     else:
                         # Check what GPU flavours are currently available using the built-in method
-                        free_flavours = list(spawner.gpus.get_free_gpu_flavours().keys())
-                        
+                        if spawner.LHCB_SWAN_ROLE in spawner.user_roles:
+                            free_flavours = list(spawner.gpus.get_free_lhcb_gpu_flavours().keys())
```
etejedor, line 131: I think I would be slightly in favour of pushing this down to the get_free_gpu method, perhaps by parameterizing it to make it receive the user roles. The get_free_gpu method should be the brains and hide this complexity from its callers.

PR #416, opened 2025-10-01
Remove unecessary code for GPU allocation

PMax5:


swan-cern/files/swan_computing_config.py
```diff
@@ -144,9 +141,6 @@ async def _modify_pod_for_gpu(self):
                 raise
             except Exception as e:
                 spawner.log.error(f'Error updating free GPU count for {gpu_description}: {e}') # Don't fail pod creation if tracking fails
-            
-            # The GPU flavour requested by the user is available, proceed with user pod creation
```
etejedor, line 148: Why removing this? We need the gpu_resource_name that comes from the form, and this is it, right?

etejedor (approved):

PR #501, opened 2026-04-07
Make it possible to run with EOS disabled

tomasr8:
For the EOSC node, since users might have a CERN account, we do not want to expose EOS (and users might not have access to it anyway without a regular CERN account). We should make it possible to run SWAN without EOS. At least for now, this means no persistent storage at all.

This is basically the minimum set of changes that let's us spawn a session without EOS, though we might still a need a change or two in the start up scripts in jupyter-images and the kube spawner.

There was already `custom.eos.enabled` so I reused that. In addition, we also already have the `local_home` trait, which sets `HOME` to `/home/<user>` instead of EOS home.

### DIFF
--- a/swan-cern/files/swan_config_cern.py
+++ b/swan-cern/files/swan_config_cern.py
@@ -29,19 +29,15 @@ class SwanPodHookHandlerProd(SwanPodHookHandler):
     async def get_swan_user_pod(self):
         super().get_swan_user_pod()
 
-        # ATTENTION Spark requires this side container, so we need to create it!!
-        # Check if we should add the EOS path in the firstplace
-        # if hasattr(self.spawner, 'local_home') and \
-        #     not self.spawner.local_home:
-
         # Check if the user has access to the Technical Network
         self._check_tn_access()
 
-        # get eos token
-        eos_secret_name = await self._init_eos_secret()
-
-        # init user containers (notebook and side-container)
-        self._init_eos_containers(eos_secret_name)
+        # When eos is enabled, create eos token and side-container for token refresh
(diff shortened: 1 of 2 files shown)


swan-cern/files/swan_config_cern.py
```diff
@@ -29,19 +29,15 @@ class SwanPodHookHandlerProd(SwanPodHookHandler):
...
-        # get eos token
-        eos_secret_name = await self._init_eos_secret()
-
-        # init user containers (notebook and side-container)
-        self._init_eos_containers(eos_secret_name)
+        # When eos is enabled, create eos token and side-container for token refresh
+        # Note: Spark also requires the side container so Spark is disabled when EOS is disabled.
+        eos_enabled = get_config("custom.eos.enabled", False)
+        if eos_enabled:
+            eos_secret_name = await self._init_eos_secret()
```
tomasr8: With EOS disabled, we don't need the side container at all. This does break Spark, but we don't really need it for EOSC anyway.

swan/files/swan_config.py
```diff
@@ -141,9 +146,9 @@ def swan_pod_hook(spawner, pod):
         )
     )
 else:
-    # No access to EOS provided
-    logging.warn("EOS access not provided. Make sure you use a scratch space as home directory (local_home: true)")
-    pass
+    # No access to EOS provided, congfigure HOME to point to /home/<user> instead of EOS HOME.
+    c.SwanKubeSpawner.local_home = True
+    c.SpawnHandlersConfigs.local_home = True
```
tomasr8: This is just to explicitly set `local_home` as that's what we want when eos is disabled.

rodrigo-sobral (approved):

PR #534, opened 2026-05-06
ci: Address all code scan reports

rodrigo-sobral:


.github/workflows/release-charts.yml
```diff
@@ -101,8 +108,11 @@ jobs:
     # Update 'swan-cern' chart to use as dependency the new tagged version of 'swan' chart
     needs: [tag_swan]
     runs-on: ubuntu-latest
+    environment: bump-dependency
```
tomasr8, line 111: Why do we need this?

.github/workflows/tag-chart.yml
```diff
@@ -33,26 +30,33 @@ on:
...
       new_tag: ${{ steps.version.outputs.new_tag }}
       last_tag: ${{ steps.version.outputs.last_tag }}
+    permissions:
+      contents: write # Required to push commits and tags to master branch
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false # To use a different token with admin privileges instead of GITHUB_TOKEN in later steps
-          ref: ${{ github.ref_name }} # Checkout latest commit on branch including new commits created in earlier steps
```
tomasr8, line 47: Why was this removed?

.github/workflows/tag-chart.yml
```diff
@@ -97,11 +106,13 @@ jobs:
       NEW_TAG: ${{ needs.publish_chart.outputs.new_tag }}
       LAST_TAG: ${{ needs.publish_chart.outputs.last_tag }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        env:
+          REF_NAME: ${{ github.ref_name }}
         with:
           fetch-depth: 0 # Fetch all tags and history to check for changes
-          ref: ${{ github.ref_name }}
```
tomasr8, line 103: same here, is it because it's the default?

.github/workflows/tag-chart.yml
```diff
@@ -97,11 +106,13 @@ jobs:
...
+          persist-credentials: false
```
tomasr8, line 114: Does the `gh release` command below need the credentials from the checkout action?

.github/workflows/update-images.yml
```diff
@@ -10,32 +10,41 @@ on:
...
+    permissions:
+      contents: write # Required to push commits to master branch
+      pull-requests: write # Required to create pull requests
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         ref: master
+        persist-credentials: false
```
tomasr8, line 33: Will this work with the `git push` etc.. below?

.github/workflows/release-charts.yml
rodrigo-sobral: https://docs.zizmor.sh/audits/#secrets-outside-env

.github/workflows/tag-chart.yml
rodrigo-sobral: This got adapted to follow the same approach we do in other situations like https://github.com/swan-cern/sparkmonitor/blob/master/.github/workflows/release.yml#L40

Also, zizmor adivces to use environment variables instead (https://docs.zizmor.sh/audits/#template-injection)

.github/workflows/tag-chart.yml
rodrigo-sobral: I did a research on this `persist-credentials` in order to make PRs and releases in latter steps, and I saw that we could run this command `git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${{ github.repository }}"` for making authenticated PRs but still keeping the `persist-credentials` set to `false`. What you think about implementing that here and also in the other workflows?

.github/workflows/release-charts.yml
tomasr8: Ah it's because of the `WORKFLOW_ACCESS_TOKEN`?

.github/workflows/tag-chart.yml
tomasr8: I didn't notice `REF_NAME` above 👍

.github/workflows/tag-chart.yml
tomasr8: I think that would be fine, but will it work with the `gh` cli as well? Or does it use the `GH_TOKEN` env var?

.github/workflows/release-charts.yml
rodrigo-sobral: Exactly, the secrets are recommended to be used within an environment context

.github/workflows/tag-chart.yml
rodrigo-sobral: Yes, gh uses `GH_TOKEN` directly from the environment, it has nothing to do with git's credential helper. Since we already have `GH_TOKEN` in that step's env, `gh pr create` will work exactly as before, completely unaffected by `persist-credentials: false`.

.github/workflows/update-images.yml
rodrigo-sobral: For the same reason as discussed above, I believe so.

tomasr8 (approved):

